PR #189, opened 2024-10-15
Filter for current school when fetching users or groups

Maddi02:
This PR introduces a feature that restricts users' access to specific classes and printers based on their assigned school. The changes ensure that users can only view and interact with resources that are available to their associated school.

### Changed
- Store users by school in cache
- Store groups by school in cache
- Store schools in cache
- Global admin can fetch all users/groups
- Fix group fetch 
  - Decrease the console errors
  - Retry on failed fetches
  - Don't block API start up
 

OLD:
> It is important to clarify if handling multiple school assignments per teacher is a relevant use case, as this scenario is not supported in the current implementation.


TomlDev (changes requested):
Looks good 👍🏽

TomlDev (changes requested):
Some changes have to be reverted, others should be added 👍🏽

apps/api/src/groups/groups.service.ts
```diff
@@ -105,9 +105,9 @@ class GroupsService {
     return response.data;
   }
 
-  static async fetchAllUsers(token: string): Promise<LDAPUser[]> {
+  static async fetchAllUsers(token: string): Promise<Record<string, LDAPUser[]>> {
     try {
-      return await GroupsService.makeAuthorizedRequest<LDAPUser[]>(HttpMethods.GET, 'users', token);
+      return await GroupsService.makeAuthorizedRequest<Record<string, LDAPUser[]>>(HttpMethods.GET, 'users', token);
```
TomlDev, line 110: I think this did not change.
When you query keycloak it does return an array of users and not an object with `[key]: LDAPUser[]`

apps/api/src/users/users.service.ts
```diff
@@ -53,21 +53,38 @@ class UsersService {
...
+          acc[school].push(user);
+        }
+        return acc;
+      },
+      {} as Record<string, LDAPUser[]>,
+    );
+
+    await this.cacheManager.set('allUsersBySchool', usersBySchool, DEFAULT_CACHE_TTL_MS);
+
+    return usersBySchool;
```
TomlDev, line 81: This gives an error when I search for a user in create conference:
grafik

libs/src/groups/types/ldapUserAttributes.ts
```diff
@@ -2,6 +2,7 @@
 // Any modifications should be carefully reviewed to ensure compatibility with the source.
 export type LDAPUserAttributes = {
   LDAP_ENTRY_DN: string[];
+  school: string[];
```
TomlDev, line 5: When I checked the `fetchAllUsers` response the LDAPUser was only this, are you sure that these attributes need an adjustment?
```
  {
    "id": "527b810e-19ef-434e-bf2c-1e274e57be32",
    "username": "johnsomi",
    "firstName": "Min-Jun",
    "lastName": "Johnson",
    "email": "johnsomi@lm.schulung.multi.schule",
    "emailVerified": true,
    "createdTimestamp": 1729860502977,
    "enabled": true,
    "federationLink": "30f7a925-f6c4-424c-bf83-88ea3acf68d3"
  },

File: apps/api/src/lmnApi/lmnApi.service.ts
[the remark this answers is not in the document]
Comment (TomlDev) [reply apps/api/src/lmnApi/lmnApi.service.ts:1]: Alright, yes if you don't find a solution we can try to debug it together 👍🏽

File: libs/src/groups/types/ldapUserAttributes.ts (hunk 3 above)
Comment (TomlDev) [reply libs/src/groups/types/ldapUserAttributes.ts:5]: `school` must be optional since it's not always present

File: apps/api/src/users/users.service.ts [hunk 4]
```diff
@@ -53,21 +52,37 @@ class UsersService {
     return result.deletedCount > 0;
   }
 
-  async findAllCachedUsers(token: string): Promise<LDAPUser[]> {
-    const cachedUsers = await this.cacheManager.get<LDAPUser[]>('allUsers');
-    if (cachedUsers) {
-      return cachedUsers;
+  async findAllCachedUsers(token: string, school: string): Promise<Record<string, LDAPUser[]>> {
```
TomlDev, line 55: Your changes are valid and OK 👍🏽 

But I think we have an opportunity here to improve it further:
Redis is used as a key/value database and therefore we can now divide the users entry by schools.

Before we had one BIG entry for all users but now we can have a smaller entry for each school.
So what we have to do is:

1. Fetch all users (Keycloak natively does not support filtering for example for `attributes.school`)
2. Sort the users by schools
3. Store them in Redis by school
```
const ALL_USERS_CACHE_KEY = 'allUsers-';  // this should be in the shared-lib
usersBySchool.forEach(school) -> 
await this.cacheManager.set(ALL_USERS_CACHE_KEY + school, usersBySchool[school], DEFAULT_CACHE_TTL_MS);
```
4. When a user fetches the users it sends his school 
```
const cachedUsers = await this.cacheManager.get<Record<string, LDAPUser[]>>(ALL_USERS_CACHE_KEY + school);
```

TomlDev (changes requested):
Some refactorings regarding the Redis store 👍🏽 
+ the tests are failing

apps/api/src/groups/groups.service.ts
michidold, line 1: I have the issue that the default school groups are not found. They follow their own pattern:
grafik

apps/api/src/groups/groups.service.ts
```diff
@@ -172,35 +188,111 @@ class GroupsService implements OnModuleInit {
...
+
+    return groups;
+  }
+
+  private async cacheSchoolGroups(groups: Group[]): Promise<Group[]> {
+    const schoolGroups = groups.filter((group) => group.path.startsWith('/s_'));
+    await this.cacheManager.set(ALL_SCHOOLS_CACHE_KEY, schoolGroups, GROUPS_CACHE_TTL_MS);
+
+    return schoolGroups;
+  }
```
michidold, line 204: Since they are used at one place, I think it makes sense to format it here as it will be needed -> `.replace('/s_', '')`

michidold (changes requested):
Found some stuff. One major issue, default school groups are not found and in classmgmt the fetch of all groups takes very long. Seems its related to the first issue

apps/api/src/groups/groups.service.ts
TomlDev: I don't think that it makes sense to manipulate the original data since we have to use the full name in places.
For example when sending the request to join a class, etc.

apps/api/src/groups/groups.service.ts
TomlDev: Really good catch, thank you.
I fixed it now.

For the long loading: As discussed today this seems to be related to LMN changes.

TomlDev (approved):
Very good and professional 👍🏽

PR #197, opened 2024-10-17
Implement the hand out feature of LMN

Maddi02:
### Description

This merge request introduces a feature that allows teachers to handout files to the students. 

### Important for Review

- **500 Error with Copy Command**: The `COPY` command currently triggers a 500 error, even though it successfully copies the files. This issue is being investigated in collaboration with Arnaud (see Issue [#217]) => solved ✅
  
- **No Exam Mode Functionality**: This merge request does not include functionality for Exam Mode

Subissue
#264 
#620


apps/api/src/filesharing/filesharing.controller.ts
```diff
@@ -137,11 +138,17 @@ class FilesharingController {
...
     @GetCurrentUsername() username: string,
   ) {
     const { collectFileRequestDTO } = body;
-    return this.filesharingService.collectFiles(username, collectFileRequestDTO, userRole);
+    return this.filesharingService.collectFiles(
+      username,
+      collectFileRequestDTO,
+      userRole,
+      type as LmnApiCollectOperations,
+    );
```
TomlDev, line 151:
```
  @Post(FileSharingApiEndpoints.COLLECT)
  async collectFiles(
    @Body() body: { collectFileRequestDTO: CollectFileRequestDTO[] },
    @Query('type') type: LmnApiCollectOperations,
    @Query('userRole') userRole: string,
    @GetCurrentUsername() username: string,
  ) {
    const { collectFileRequestDTO } = body;
    return this.filesharingService.collectFiles(
      username,
      collectFileRequestDTO,
      userRole,
      type,
    );
  }

File: apps/api/src/filesharing/filesharing.service.ts [hunk 2]
```diff
@@ -342,7 +388,14 @@ class FilesharingService {
...
       try {
-        await this.moveOrRenameResource(username, item.originPath, item.destinationPath);
+        if (type === LmnApiCollectOperations.CUT) {
+          await this.cuteCollectedItems(username, item.originPath, item.destinationPath);
+        } else {
+          await this.copyCollectedItems(username, {
+            originFilePath: item.originPath,
+            destinationFilePaths: [`${item.destinationPath}/${getCurrentTimestamp()}`],
+          });
+        }
```
TomlDev, line 398: The `collectFiles` fn has many catch blocks that only log to the console. Does that make sense?
If it fails to create the initial folder, does it make sense to continue with the later operations?

apps/api/src/filesharing/filesharing.service.ts
```diff
@@ -262,26 +264,67 @@ class FilesharingService {
     );
   };
 
-  async duplicateFile(username: string, duplicateFile: DuplicateFileRequestDto) {
-    const client = await this.getClient(username);
+  private async createFolderIfNotExists(username: string, destinationPath: string) {
+    const sanitizedDestinationPath = destinationPath.replace(`${FILE_PATHS.COLLECT}/`, '');
```
TomlDev, line 268: I think since the function uses this hardcoded `FILE_PATHS.COLLECT` replace I wouldn't give it a general name like `createFolderIfNotExists`

apps/api/src/filesharing/filesharing.service.ts
```diff
@@ -262,26 +264,67 @@ class FilesharingService {
...
+        FileSharingErrorMessage.CreationFailed,
+        HttpStatus.NOT_FOUND,
+        error,
+        FilesharingService.name,
+      );
+    }
+  }
 
+  private async copyFile(client: AxiosInstance, originPath: string, destinationPath: string) {
+    const sanitizedDestinationPath = destinationPath.replace(/\u202F/g, ' ');
```
TomlDev, line 283: Why is this character in the path?

apps/api/src/filesharing/filesharing.service.ts
```diff
@@ -262,26 +264,67 @@ class FilesharingService {
...
+      throw new CustomHttpException(
+        FileSharingErrorMessage.CreationFailed,
+        HttpStatus.NOT_FOUND,
+        error,
+        FilesharingService.name,
+      );
+    }
+  }
 
+  private async copyFile(client: AxiosInstance, originPath: string, destinationPath: string) {
```
TomlDev, line 282: The linter says "ESLint: Expected 'this' to be used by class async method 'copyFile'.(@typescript-eslint/class-methods-use-this)"

apps/api/src/filesharing/filesharing.service.ts
```diff
@@ -262,26 +264,67 @@ class FilesharingService {
...
+    const sanitizedDestinationPath = destinationPath.replace(/\u202F/g, ' ');
+    try {
+      await FilesharingService.copyFileViaWebDAV(client, encodeURI(originPath), sanitizedDestinationPath);
+    } catch (error) {
+      Logger.log(error); // TODO: Replace this with a custom exception. Issue #217
+    }
+  }
+
+  cuteCollectedItems = async (username: string, originPath: string, newPath: string): Promise<WebdavStatusReplay> => {
```
TomlDev, line 291: to cut /= cute

apps/frontend/src/components/ui/Dialog.tsx
```diff
@@ -76,9 +76,18 @@ const DialogHeader = ({ className, ...props }: React.HTMLAttributes<HTMLDivEleme
 );
 DialogHeader.displayName = 'DialogHeader';
 
-const DialogFooter = ({ className, ...props }: React.HTMLAttributes<HTMLDivElement>) => (
+interface DialogFooterProps extends React.HTMLAttributes<HTMLDivElement> {
+  layout?: 'one-column' | 'two-column';
```
TomlDev, line 80: Since it's used multiple times, please move it to the shared lib.
Also normally I would suggest to use an enum instead then. Or as we now do, use a const

apps/frontend/src/locales/de/translation.json
```diff
@@ -293,7 +293,10 @@
     "saveSession": "Sitzung speichern",
     "myRoom": "Mein Raum",
     "sharedMailBox": "Geteiltes Postfach",
-    "enrolPageDescription": "Wählen Sie die Klassen und Projekte aus, für die Sie sich einschreiben möchten, indem Sie die Checkboxen aktivieren."
+    "enrolPageDescription": "Wählen Sie die Klassen und Projekte aus, für die Sie sich einschreiben möchten, indem Sie die Checkboxen aktivieren.",
```
TomlDev, line 296: du

apps/frontend/src/locales/de/translation.json
```diff
@@ -293,7 +293,10 @@
...
+    "collectAndCut": "Ausschneiden",
+    "collectAndCopy": "Kopieren",
+    "copyOrCut": "Möchten Sie die Dateien ausschneiden oder kopieren?"
```
TomlDev, line 299: du

apps/frontend/src/pages/ClassManagement/LessonPage/useLessonStore.ts
```diff
@@ -14,6 +14,7 @@ import { HTTP_HEADERS } from '@libs/common/types/http-methods';
 import DuplicateFileRequestDto from '@libs/filesharing/types/DuplicateFileRequestDto';
 import CollectFileRequestDTO from '@libs/filesharing/types/CollectFileRequestDTO';
 import FileSharingApiEndpoints from '@libs/filesharing/types/fileSharingApiEndpoints';
+import LmnAp from '@libs/lmnApi/types/lmnApiCollectOperations';
```
TomlDev, line 17: Better import as `LmnApiCollectOperations`

libs/src/classManagement/types/store/lessonStore.ts
```diff
@@ -5,6 +5,7 @@ import LmnApiSchoolClass from '@libs/lmnApi/types/lmnApiSchoolClass';
 import UserLmnInfo from '@libs/lmnApi/types/userInfo';
 import DuplicateFileRequestDto from '@libs/filesharing/types/DuplicateFileRequestDto';
 import CollectFileRequestDTO from '@libs/filesharing/types/CollectFileRequestDTO';
+import LmnAp from '@libs/lmnApi/types/lmnApiCollectOperations';
```
TomlDev, line 8: LmnApiCollectOperations

apps/frontend/src/pages/ClassManagement/components/Dialogs/CollectFilesDialog.tsx
TomlDev, line 1: Something to the "share" dialog:
image

image

apps/frontend/src/pages/ClassManagement/components/Dialogs/CollectFilesDialog.tsx
```diff
@@ -1,22 +1,84 @@
...
+
+      <div className="absolute bottom-0 left-0 right-0 flex justify-end p-4">
+        <Button
+          type="button"
+          size="lg"
+          variant="btn-collaboration"
+          onClick={action}
+        >
+          {t(`classmanagement.${title}`)}
+        </Button>
```
TomlDev, line 80: Currently when you click "collect" it loads, then it finish loading but there is no acknowledgment to the user if it worked. There should be a toaster

apps/frontend/src/pages/ClassManagement/components/Dialogs/CollectFilesDialog.tsx
```diff
@@ -1,22 +1,84 @@
...
+              <DropdownMenuItem
+                key={option.id}
+                icon={option.icon}
+                onClick={() => setSelectedOption(option)}
+              >
+                {option.label}
+              </DropdownMenuItem>
+            ))}
+          </DropdownMenuContent>
+        </DropdownMenuSH>
```
TomlDev, line 69: image
image

apps/frontend/src/pages/ClassManagement/components/Dialogs/CollectFilesDialog.tsx
TomlDev, line 1: There is something going wrong when switching the dialogs. Either it should show a loading icon or be empty until data is loaded

https://github.com/user-attachments/assets/144d428e-96ea-45b5-8cf6-404306de31d6

apps/frontend/src/pages/ClassManagement/components/Dialogs/CollectFilesDialog.tsx
```diff
@@ -1,22 +1,84 @@
...
+    },
+    {
+      id: '2',
+      label: t('classmanagement.collectAndCopy'),
+      icon: <FaCopy />,
+      action: () => setCollectionType(LmnApiCollectOperations.COPY),
+    },
+  ];
+
+  const [selectedOption, setSelectedOption] = useState(options[0]);
```
TomlDev, line 36: There is normally no need to handle a state locally in a component AND in the store 
For example you could use 
- an array like yours with `id: LmnApiCollectOperations`:
```
const selectedOption = options.find(o => o.id === useLessonStore().selectedCollectionType)
```

- an object
```
const options = {
  [LmnApiCollectOperations.CUT]: {
      label: t('classmanagement.collectAndCut'),
      icon: <FaCut />,
  }
}
```

and access it with 
`const selectedOption =  options[useLessonStore().selectedCollectionType]`

apps/frontend/src/pages/ClassManagement/components/Dialogs/CollectFilesDialog.tsx
```diff
@@ -1,22 +1,84 @@
...
-    <div className="mt-4 flex justify-between space-x-4">
-      <Button
-        type="button"
-        size="lg"
-        variant="btn-collaboration"
-        onClick={action}
-      >
-        {t(`classmanagement.${title}`)}
-      </Button>
+    <div className="h-[120px] justify-between py-2">
```
TomlDev, line 47: I don't think that this belongs into the footer.
It should be part of the body because it's the content of the dialog.
The AI:

> In a React dialog component, the **footer** serves as a designated area at the bottom of the dialog for user actions or additional information. Common uses include:
> 
> 1. **Action Buttons**: Frequently, the footer contains buttons for actions such as "Save," "Cancel," or "Confirm." These buttons enable users to proceed with or dismiss the dialog's purpose without additional navigation.
> 2. **Status Information**: A footer can also display supplementary information, such as form submission status, validation messages, or loading indicators, helping users understand the dialog's state or any required next steps.
> 3. **Additional Context**: Some footers offer helpful context or tips related to the dialog content, making it easier for users to make informed decisions or understand any dependencies.
> 
> In general, a footer helps establish a clear layout and supports better usability by keeping actions accessible and organized, especially in dialogs with complex or multi-step content. It also reinforces visual hierarchy by clearly separating actions from the main content.

apps/frontend/src/components/ui/DropdownMenuSH.tsx
```diff
@@ -60,19 +78,18 @@ const DropdownMenuContent = React.forwardRef<
...
-      className={cn(
-        'z-10 min-w-[8rem] overflow-hidden rounded-xl border bg-popover bg-white p-1 text-popover-foreground shadow-md',
-        'transition duration-700 ease-in-out',
-        'data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=open]:fade-in-0 data-[state=closed]:zoom-out-95 data-[state=open]:zoom-in-95 data-[side=bottom]:slide-in-from-top-2 data-[side=left]:slide-in-from-right-2 data-[side=right]:slide-in-from-left-2 data-[side=top]:slide-in-from-bottom-2',
-        className,
-      )}
-      {...props}
-    />
-  </DropdownMenuPrimitive.Portal>
+  <DropdownMenuPrimitive.Content
```
TomlDev, line 81: Why was the portal removed?

TomlDev (changes requested):
👍🏽

apps/api/src/filesharing/filesharing.service.ts
```diff
@@ -262,34 +264,80 @@ class FilesharingService {
...
+    }
+  }
 
-    const fullOriginPath = `${this.baseurl}${duplicateFile.originFilePath}`;
+  private static async copyFile(client: AxiosInstance, originPath: string, destinationPath: string) {
+    const sanitizedDestinationPath = destinationPath.replace(/\u202F/g, ' ');
+    try {
+      await FilesharingService.copyFileViaWebDAV(client, encodeURI(originPath), sanitizedDestinationPath);
+    } catch (error) {
+      Logger.log(error); // TODO: Replace this with a custom exception. Issue #217
```
TomlDev, line 287: TODO

apps/api/src/filesharing/filesharing.service.ts
```diff
@@ -381,7 +483,7 @@ class FilesharingService {
       },
       FileSharingErrorMessage.DuplicateFailed,
       (response: WebdavStatusReplay) => ({
-        success: response.status >= 200 && response.status < 300,
+        success: response.success,
```
TomlDev, line 486: This transformer does not seem to be necessary

apps/api/src/lmnApi/lmnApi.service.ts
```diff
@@ -682,6 +691,24 @@ class LmnApiService {
...
+        return this.fileSharingService.createFolder(member.name, unixPath, teacher);
+      }),
+    );
+  }
+
+  private static convertWindowsToUnixPath(windowsPath: string): string {
+    const pathWithoutServer = windowsPath.replace(/^\\\\[^\\]+\\/, '');
+    const pathWithoutFirstElement = pathWithoutServer.split('\\').slice(1).join('/');
+    return `${pathWithoutFirstElement}/transfer`;
+  }
```
TomlDev, line 711: This seems to be a utility function -> shared-lib

apps/frontend/src/components/ui/AdaptiveDialog.tsx
TomlDev, line 1: Can you explain the undocumented changes in this file?
Please also add screenshots or a video to demonstrate it

apps/frontend/src/components/ui/Dialog.tsx
TomlDev, line 1: Can you provide some screenshots for this unrelated and undocumented change?

apps/frontend/src/components/ui/DropdownMenuSH.tsx
TomlDev, line 1: I tried to understand where these changes are used, but couldn't find it.

apps/frontend/src/components/ui/DropdownMenuSH.tsx
```diff
@@ -20,7 +20,25 @@ import cn from '@libs/common/utils/className';
...
+  <DropdownMenuPrimitive.Trigger
+    ref={ref}
+    className={cn('flex items-center gap-2', className)}
+    {...props}
+  >
+    <div className="flex items-center gap-2">
+      {selectedOptionIcon && <span>{selectedOptionIcon}</span>}
+      {selectedOptionLabel ? <span>{selectedOptionLabel}</span> : children}
+    </div>
+  </DropdownMenuPrimitive.Trigger>
```
TomlDev, line 39: I am unsure what the goal of this is. Normally I would expect that anything can be passed as Trigger, including icons and labels so there is no necessity to add this logic here

apps/frontend/src/pages/ClassManagement/LessonPage/useLessonStore.ts
```diff
@@ -86,20 +92,26 @@ const useLessonStore = create<LessonStore>(
         } catch (error) {
           handleApiError(error, set);
         } finally {
+          toast.success(t('classmanagement.filesShared'));
```
TomlDev, line 95: This will show the success message in every case, also in the catch (error) case. Move it inside the try block

apps/frontend/src/pages/ClassManagement/LessonPage/useLessonStore.ts
```diff
@@ -86,20 +92,26 @@ const useLessonStore = create<LessonStore>(
...
-        const queryParamString = `?userRole=${encodeURIComponent(userRole)}`;
+        const queryParamString = `?type=${type}&userRole=${userRole}`;
         try {
           await eduApi.post(`${FileSharingApiEndpoints.BASE}/${FileSharingApiEndpoints.COLLECT}/${queryParamString}`, {
             collectFileRequestDTO,
           });
         } catch (error) {
           handleApiError(error, set);
         } finally {
+          toast.success(t('classmanagement.filesCollected'));
```
TomlDev, line 114: Same as above

apps/frontend/src/pages/ClassManagement/components/Dialogs/CollectFilesDialog.tsx
TomlDev, line 1: grafik

apps/frontend/src/pages/ClassManagement/components/Dialogs/CollectFilesDialog.tsx
```diff
@@ -13,14 +13,78 @@
...
+import LMN_API_COLLECT_OPERATIONS from '@libs/lmnApi/constants/lmnApiCollectOperations';
 
 const CollectFilesDialog: React.FC<ShareCollectDialogProps> = ({ title, isOpen, onClose, action }) => {
-  const getDialogBody = () => <div className="text-background">{t('classmanagement.CollectFilesDescription')}</div>;
+  const { collectionType, setCollectionType } = useLessonStore();
+
+  const options: Record<
+    LmnApiCollectOperationsType,
+    {
+      id: LmnApiCollectOperationsType;
```
TomlDev, line 31: This seems strange, either use it as key in an object or as id in an array

apps/frontend/src/pages/FileSharing/dialog/DialogBodys/MoveContentDialogWrapper.tsx
```diff
@@ -0,0 +1,32 @@
...
+ * You should have received a copy of the GNU Affero General Public License along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+
+import React from 'react';
+import useUserPath from '@/pages/FileSharing/hooks/useUserPath';
+import MoveContentDialogBody from '@/pages/FileSharing/dialog/DialogBodys/MoveContentDialogBody';
+import MoveContentDialogProps from '@libs/filesharing/types/moveContentDialogProps';
+import ContentType from '@libs/filesharing/types/contentType';
+
+const MoveContentDialogWrapper: React.FC<Omit<MoveContentDialogProps, 'pathToFetch'>> = (props) => {
```
TomlDev, line 19: Instead of `wrapper` which has virtually no meaning please find a better, descriptive name.
For what is it specially used? Or in which place?

apps/frontend/src/pages/FileSharing/table/FileSharingTable.tsx
```diff
@@ -39,7 +39,7 @@ const FileSharingTable = () => {
 
   return (
     <ScrollableTable
-      columns={FileSharingTableColumns}
+      columns={FileSharingTableColumns()}
```
TomlDev, line 42: If this is a function it violates some spelling rules

apps/frontend/src/pages/FileSharing/dialog/DialogBodys/MoveContentDialogBody.tsx
TomlDev, line 1: Please check if you can adapt some of the changes the AI proposes:

> Below is a **refactored** version of the same component. It preserves all **logic** and **behavior** but makes the code slightly more readable and consistent:
> 
> - **Unifies** the `useEffect` logic for fetching data (fewer nested `if`s).  
> - **Slightly simplifies** the row‑selection handler.  
> - **Clarifies** variable names in `onFilenameClick`.  
> - **Adjusts** minor stylistic details in `getHiddenSegments`.  
> 
> Everything else (function signatures, data flow, etc.) remains **unchanged**.
> 
> ```tsx
> /*
>  * LICENSE
>  *
>  * This program is free software: you can redistribute it and/or modify it under the terms of the GNU Affero General Public License as published by
>  * the Free Software Foundation, either version 3 of the License, or (at your option) any later version.
>  *
>  * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
>  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more details.
>  *
>  * You should have received a copy of the GNU Affero General Public License along with this program. If not, see <https://www.gnu.org/licenses/>.
>  */
> 
> import { DirectoryFileDTO } from '@libs/filesharing/types/directoryFileDTO';
> import React, { useEffect, useState } from 'react';
> import { useTranslation } from 'react-i18next';
> import useFileSharingStore from '@/pages/FileSharing/useFileSharingStore';
> import DirectoryBreadcrumb from '@/pages/FileSharing/breadcrumb/DirectoryBreadcrumb';
> import useFileSharingDialogStore from '@/pages/FileSharing/dialog/useFileSharingDialogStore';
> import ScrollableTable from '@/components/ui/Table/ScrollableTable';
> import APPS from '@libs/appconfig/constants/apps';
> import { FileSharingTableColumns } from '@/pages/FileSharing/table/FileSharingTableColumns';
> import { ColumnDef, OnChangeFn, RowSelectionState } from '@tanstack/react-table';
> import FILESHARING_TABLE_COLUM_NAMES from '@libs/filesharing/constants/filesharingTableColumNames';
> import MoveContentDialogBodyProps from '@libs/filesharing/types/moveContentDialogProps';
> import ContentType from '@libs/filesharing/types/contentType';
> import LoadingIndicator from '@/components/shared/LoadingIndicator';
> import useLmnApiStore from '@/store/useLmnApiStore';
> 
> const MoveContentDialogBody: React.FC<MoveContentDialogBodyProps> = ({
>   showAllFiles = false,
>   pathToFetch,
>   showSelectedFile = true,
>   showHome = true,
>   fileType,
> }) => {
>   const { t } = useTranslation();
>   const [currentPath, setCurrentPath] = useState(pathToFetch || '');
> 
>   const { user } = useLmnApiStore();
> 
>   const {
>     setMoveOrCopyItemToPath,
>     moveOrCopyItemToPath,
>   } = useFileSharingDialogStore();
> 
>   const {
>     fetchDialogDirs,
>     fetchDialogFiles,
>     dialogShownDirs,
>     dialogShownFiles,
>     isLoading,
>   } = useFileSharingStore();
> 
>   // Decide which fetch function to use based on fileType
>   const fetchMechanism =
>     fileType === ContentType.DIRECTORY
>       ? fetchDialogDirs
>       : fetchDialogFiles;
> 
>   // Decide which files/dirs to show based on fileType
>   const files =
>     fileType === ContentType.DIRECTORY
>       ? dialogShownDirs
>       : dialogShownFiles;
> 
>   // Handle row selection in the ScrollableTable
>   const handleRowSelectionChange: OnChangeFn<RowSelectionState> = (updaterOrValue) => {
>     const selectionValue =
>       typeof updaterOrValue === 'function' ? updaterOrValue({}) : updaterOrValue;
> 
>     const selectedItems = Object.entries(selectionValue)
>       .filter(([_, isSelected]) => isSelected)
>       .map(([rowId]) => files.find((file) => file.filename === rowId))
>       .filter(Boolean) as DirectoryFileDTO[];
> 
>     // Only care about the first selected item
>     setMoveOrCopyItemToPath(selectedItems[0]);
>   };
> 
>   // Clicking a directory row should navigate into that directory
>   const onFilenameClick = (item: DirectoryFileDTO) => {
>     if (item.type === ContentType.DIRECTORY) {
>       let newPath = currentPath;
>       if (!newPath.endsWith('/')) {
>         newPath += '/';
>       }
>       if (newPath === '/') {
>         newPath += item.filename
>           .replace('/webdav/', '')
>           .replace(`server/${user?.school}/`, '');
>       } else {
>         newPath += item.basename;
>       }
>       setCurrentPath(newPath);
>     }
>   };
> 
>   // Fetch logic on mount/update
>   useEffect(() => {
>     // If we don't want to show all files, simply fetch currentPath
>     if (!showAllFiles) {
>       void fetchMechanism(currentPath);
>       return;
>     }
> 
>     // showAllFiles is true
>     // 1) If pathToFetch is undefined => fetch currentPath
>     if (!pathToFetch) {
>       void fetchMechanism(currentPath);
>       return;
>     }
> 
>     // 2) pathToFetch is defined
>     // if currentPath includes pathToFetch => fetch currentPath
>     // else => fetch pathToFetch
>     if (currentPath.includes(pathToFetch)) {
>       void fetchMechanism(currentPath);
>     } else {
>       void fetchMechanism(pathToFetch);
>     }
> 
>     // eslint-disable-next-line react-hooks/exhaustive-deps
>   }, [currentPath]); // Keep original dependency to match the existing logic
> 
>   // Allows the DirectoryBreadcrumb to navigate "up" or to specific subpaths
>   const handleBreadcrumbNavigate = (path: string) => {
>     const newPath = path.replace('webdav/', '');
>     setCurrentPath(newPath);
>   };
> 
>   // Gather hidden segments from pathToFetch
>   const getHiddenSegments = (): string[] => {
>     if (!pathToFetch) return [];
>     const segments = pathToFetch.split('/');
>     // The original code tries to find the last segment index
>     const index = segments.findIndex(
>       (segment) => segment === segments.at(segments.length - 1)
>     );
>     return index > -1 ? segments.slice(0, index) : [];
>   };
> 
>   // Footer element: show selected item or a message to select a file
>   const footer = (
>     <div className="bottom-0 justify-end bg-gray-100 p-4 text-sm text-foreground">
>       {moveOrCopyItemToPath?.basename && showSelectedFile ? (
>         <p className="bg-gray-100">
>           {t('moveItemDialog.selectedItem')}: {moveOrCopyItemToPath.basename}
>         </p>
>       ) : (
>         <p className="bg-gray-100">
>           <span>{t('filesharing.selectFile')}</span>
>         </p>
>       )}
>     </div>
>   );
> 
>   // Only show a single "filename with selection checkbox" column
>   const visibleColumns = [FILESHARING_TABLE_COLUM_NAMES.SELECT_FILENAME];
>   const columns: ColumnDef<DirectoryFileDTO>[] = FileSharingTableColumns(
>     visibleColumns,
>     onFilenameClick
>   );
> 
>   return (
>     <div className="h-[60vh] flex-col overflow-auto text-background scrollbar-thin">
>       <LoadingIndicator isOpen={isLoading} />
>       <div className="pb-2">
>         <DirectoryBreadcrumb
>           path={currentPath}
>           onNavigate={handleBreadcrumbNavigate}
>           showHome={showHome}
>           hiddenSegments={getHiddenSegments()}
>           showTitle={false}
>         />
>       </div>
>       {!isLoading && (
>         <div className="overflow-auto">
>           <ScrollableTable
>             columns={columns}
>             data={files}
>             selectedRows={
>               moveOrCopyItemToPath
>                 ? { [moveOrCopyItemToPath.filename]: true }
>                 : {}
>             }
>             onRowSelectionChange={handleRowSelectionChange}
>             applicationName={APPS.FILE_SHARING}
>             getRowId={(row) => row.filename}
>             showHeader={false}
>             textColorClass="text-background"
>             showSelectedCount={false}
>             footer={footer}
>             filterKey="select-filename"
>             filterPlaceHolderText="filesharing.filterPlaceHolderText"
>           />
>         </div>
>       )}
>     </div>
>   );
> };
> 
> export default MoveContentDialogBody;
> ```
> 
> ## What Changed (No Logic Altered)
> 
> 1. **`useEffect`**  
>    - Combined the nested `if` statements into a clearer hierarchy:  
>      - If `!showAllFiles` → fetch `currentPath`.  
>      - Else if no `pathToFetch` → fetch `currentPath`.  
>      - Else check if `currentPath` includes `pathToFetch`.  
>    - Preserves the original behavior entirely.
> 
> 2. **Selection Handler**  
>    - Rewrote the row‑selection function to use `Object.entries(...)` instead of `Object.keys(...)`, which is a tad clearer when checking for `true`/`false`. Still returns the first item, as originally.
> 
> 3. **`onFilenameClick`**  
>    - Same logic, but condensed the string manipulation into one block before calling `setCurrentPath`.
> 
> 4. **`getHiddenSegments`**  
>    - Tidied up the index lookup. The original code used `segments.at(segment.length)`—this has the same net effect as `segments.at(segments.length - 1)` for the last element.
> 
> All these changes preserve the **same flow** and **same conditions**—so nothing functionally changes. The code should behave exactly as before but is a bit more readable and consistent.

apps/frontend/src/pages/FileSharing/table/FileSharingTableColumns.tsx
```diff
@@ -29,177 +29,195 @@ import ContentType from '@libs/filesharing/types/contentType';
...
+              setCurrentlyEditingFile(null);
+              searchParams.set('path', getPathWithoutWebdav(row.original.filename));
+              setSearchParams(searchParams);
+            } else {
+              await resetCurrentlyEditingFile(row.original);
+            }
+          } else {
+            onFilenameClick(row.original);
+          }
+        };
```
TomlDev, line 83: Ai proposes:

> Below is one way to apply **early returns** to simplify the control flow. The logic remains the same, but it’s a bit clearer and easier to follow:
> 
> ```ts
> const handleFilenameClick = async () => {
>   // If an onFilenameClick handler was provided, just call it and exit.
>   if (onFilenameClick) {
>     onFilenameClick(row.original);
>     return;
>   }
> 
>   // If clicking is currently disabled, do nothing.
>   if (isCurrentlyDisabled) {
>     return;
>   }
> 
>   // Otherwise, proceed with the default behavior.
>   setShowEditor(false);
>   setPublicDownloadLink('');
> 
>   if (row.original.type === ContentType.DIRECTORY) {
>     setCurrentlyEditingFile(null);
>     searchParams.set('path', getPathWithoutWebdav(row.original.filename));
>     setSearchParams(searchParams);
>   } else {
>     // This needs `async/await`, so keep the function `async`.
>     await resetCurrentlyEditingFile(row.original);
>   }
> };
> ```
> 
> ### What Changed
> 
> 1. **Early Return**  
>    - We immediately return after calling `onFilenameClick` when `onFilenameClick` exists, so no else-block is required.
>    - If `isCurrentlyDisabled`, we return immediately as well, short-circuiting the rest.
> 
> 2. **Same Logic, Clearer Flow**  
>    - The core actions and conditions are unaltered, just reorganized. This makes it easier to see in which scenario each piece of code runs.

apps/frontend/src/pages/FileSharing/table/FileSharingTableColumns.tsx
```diff
@@ -29,177 +29,195 @@ import ContentType from '@libs/filesharing/types/contentType';
...
+          if (row.original.type === ContentType.FILE) {
+            return (
+              <FileIconComponent
+                filename={item.filename}
+                size={Number(TABLE_ICON_SIZE)}
+              />
+            );
+          }
+          return <MdFolder size={TABLE_ICON_SIZE} />;
+        };
```
TomlDev, line 102: Please extract this method to the top of the file, outside of the array

apps/frontend/src/pages/FileSharing/useFileSharingStore.ts
```diff
@@ -29,6 +29,8 @@ import { ResponseType } from '@libs/common/types/http-methods';
 
 type UseFileSharingStore = {
   files: DirectoryFileDTO[];
+  dialogShownFiles: DirectoryFileDTO[];
```
TomlDev, line 32: Please move everything related to the dialog to it's own store.

libs/src/classManagement/types/store/lessonStore.ts
```diff
@@ -26,12 +27,17 @@ interface LessonState {
   member: UserLmnInfo[];
   groupTypeFromStore: string | undefined;
   groupNameFromStore: string | undefined;
+  collectionType: LmnApiCollectOperationsType;
```
TomlDev, line 30: I think this should be moved to another store. From the place where it is defined and its name it's intent is totally unclear.
The minimum would be to rename it to something way more straight forward. `collectDialogCurrentlySelectedCollectionOperation` or something, I don't know

libs/src/filesharing/utils/getCurrentTimestamp.ts
```diff
@@ -0,0 +1,26 @@
...
+ * This program is free software: you can redistribute it and/or modify it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+
+const getCurrentTimestamp = (): string =>
```
TomlDev, line 13: Please add an example as comment so we can see what kind of timestamp this produces

apps/frontend/src/components/ui/AdaptiveDialog.tsx
```diff
@@ -83,11 +103,21 @@ const AdaptiveDialog: FC<AdaptiveDialogProps> = ({
...
         variant={variant}
-        className={desktopContentClassName}
+        className={`${desktopContentClassName} overflow-x-hidden `}
       >
         <DialogTitle>{title}</DialogTitle>
-        {body}
-        {footer ? <DialogFooter>{footer}</DialogFooter> : null}
+        <VisuallyHidden>
+          <DialogTitle>{title}</DialogTitle>
+        </VisuallyHidden>
```
TomlDev, line 111: Why this?

TomlDev (changes requested):
Please have a look at the comments.
I fixed the problem that the lesson screen did not restore the users when the app was switched.

apps/api/src/filesharing/filesharing.service.ts
```diff
@@ -266,35 +270,126 @@ class FilesharingService {
...
 
-      try {
-        await this.createFolder(username, pathWithoutFilename, FILE_PATHS.COLLECT);
-      } catch (error) {
-        Logger.error(error);
-      }
+    try {
+      await this.createCollectFolderIfNotExists(username, originPath);
+    } catch (error) {
+      Logger.error(error);
```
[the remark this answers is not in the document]
Maddi02: There is no reason. Changed it to a `CustomHttpException`

apps/frontend/src/components/ui/AdaptiveDialog.tsx
Maddi02: There is no good reason to change this component. I guess it was some stuff that should have been fixed some time ago. However, these changes are no longer needed.

apps/frontend/src/pages/ClassManagement/components/Dialogs/CollectFilesDialog.tsx
Maddi02: Should be fixed

image

libs/src/filesharing/utils/getCurrentTimestamp.ts
Maddi02: `LOG [26-02-2025_18-30-03] currentTimeStamp`

apps/frontend/src/pages/ClassManagement/components/Dialogs/CollectFilesDialog.tsx
Maddi02: image

Maddi02:
This PR contains the solution to #538. Due to the fact that moving files like `03+-+Data+Models+and+Query+Languages.pdf`to another folder does not work.

apps/api/src/filesharing/filesharing.service.ts
```diff
@@ -266,35 +270,126 @@ class FilesharingService {
...
+      const files = await this.getFilesAtPath(username, parentPath);
+      const existingFiles = new Set(
+        files.filter((item) => item.type === ContentType.FILE).map((item) => item.basename),
+      );
+      return existingFiles.has(name);
+    }
+    return false;
   }
 
+  /* eslint-disable no-await-in-loop, no-restricted-syntax */
```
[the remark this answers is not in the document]
TomlDev: Okay :D I meant to refactor so no eslint rules must be disabled.
Something like this:
```
/* eslint-disable no-await-in-loop */ // <-- remove this and see new approach
async duplicateFile(username: string, duplicateFile: DuplicateFileRequestDto): Promise<WebdavStatusResponse> {
  const client = await this.getClient(username);

  await Promise.all(
    duplicateFile.destinationFilePaths.map(async (destinationPath) => {
      const pathUpToTransferFolder = FilesharingService.getPathUntilFolder(destinationPath, FILE_PATHS.TRANSFER);
      const pathUpToTeacherFolder = FilesharingService.getPathUntilFolder(destinationPath, username);

      const userFolderExists = await this.checkIfFileOrFolderExists(
        username,
        pathUpToTransferFolder,
        username,
        ContentType.DIRECTORY,
      );

      if (!userFolderExists) {
        await this.createFolder(username, pathUpToTransferFolder, username);

        // If we just created the user folder, then check/create the collect folder too.
        const collectFolderExists = await this.checkIfFileOrFolderExists(
          username,
          pathUpToTeacherFolder,
          FILE_PATHS.COLLECT,
          ContentType.DIRECTORY,
        );
        if (!collectFolderExists) {
          await this.createFolder(username, pathUpToTeacherFolder, FILE_PATHS.COLLECT);
        }
      }

      await FilesharingService.copyFile(client, duplicateFile.originFilePath, destinationPath);
    }),
  );

  return { success: true, status: HttpStatus.OK };
}

File: apps/api/src/filesharing/filesharing.service.ts (hunk 18 above)
Comment (TomlDev) [reply apps/api/src/filesharing/filesharing.service.ts]: This is unchanged

File: libs/src/common/utils/convertWindowsToUnixPath.ts
Comment (TomlDev) [inline libs/src/common/utils/convertWindowsToUnixPath.ts:1]: I would move this to filesharing domain

File: libs/src/filesharing/utils/getCurrentTimestamp.ts (hunk 30 above)
Comment (TomlDev) [reply libs/src/filesharing/utils/getCurrentTimestamp.ts:13]: Haha sorry, I meant in the code. As annotation so you can see it when you hover the function 👍🏽

File: apps/api/src/filesharing/filesharing.service.ts (hunk 18 above)
Comment (Maddi02) [reply apps/api/src/filesharing/filesharing.service.ts]: i could reflect it

image

File: apps/api/src/filesharing/filesharing.service.ts [hunk 34]
```diff
@@ -282,15 +281,25 @@ class FilesharingService {
...
-      );
+
+    const folderAlreadyExistis = await this.checkIfFileOrFolderExists(
+      username,
+      pathWithoutFilename + '/',
+      FILE_PATHS.COLLECT,
+      ContentType.DIRECTORY,
+    );
+
+    if (!folderAlreadyExistis) {
```
TomlDev, line 292: Use early return instead

TomlDev (approved):
@michidold please have a 2nd look 👍🏽

Maddi02:
After some discussion on how to display the file-sharing progress, I’d like to show you a possible approach. Could you please give me some feedback?

image

Maddi02:
Okay, I guess I won't add any extra features in this brach. Anyway its way too big. Added a retry mechanism if the share failed on puporse. 

Feel free to have a look.

https://github.com/user-attachments/assets/0a565fe7-07ba-4219-ad8f-d9bdd432b52c

apps/api/src/filesharing/filesharing.service.ts
```diff
@@ -259,40 +271,146 @@ class FilesharingService {
...
+  };
+
+  copyCollectedItems = async (
+    username: string,
+    duplicateFile: DuplicateFileRequestDto,
+  ): Promise<WebdavStatusResponse> => {
+    const client = await this.getClient(username);
+
+    const duplicationPromises = duplicateFile.destinationFilePaths.map(async (destinationPath) => {
+      await FilesharingService.copyFile(client, duplicateFile.originFilePath, destinationPath);
```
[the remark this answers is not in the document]
michidold: Ok, I was not precise, please rework every function in this file

apps/frontend/src/locales/de/translation.json
```diff
@@ -447,7 +450,12 @@
     "editSession": "Sitzung editieren",
     "myRoom": "Mein Raum",
     "sharedMailBox": "Geteiltes Postfach",
-    "enrolPageDescription": "Wählen Sie die Klassen und Projekte aus, für die Sie sich einschreiben möchten, indem Sie die Checkboxen aktivieren.",
+    "enrolPageDescription": "Wähle die Klassen und Projekte aus, für die du dich einschreiben möchtest, indem du die Checkboxen aktivierst.",
+    "collectAndCut": "Ausschneiden",
+    "collectAndCopy": "Kopieren",
```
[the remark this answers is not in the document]
michidold: They are still in place.

apps/frontend/src/pages/FileSharing/table/FileSharingTableColumns.tsx
```diff
@@ -29,177 +29,198 @@ import ContentType from '@libs/filesharing/types/contentType';
...
+      accessorKey: FILESHARING_TABLE_COLUM_NAMES.TYPE,
+      id: FILESHARING_TABLE_COLUM_NAMES.TYPE,
+      header: function Header({ column }) {
+        return (
+          <SortableHeader<DirectoryFileDTO, unknown>
+            className={hideOnMobileClassName}
+            column={column}
+          />
+        );
+      },
```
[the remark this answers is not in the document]
michidold: Can you rework the others as well? There are more `function Cell` and `function Header ` declarations

michidold (changes requested):
Had some findings, among other things the queue implementation needs to be refactored.

apps/api/src/filesharing/filesharing.service.ts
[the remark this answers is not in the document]
Maddi02: "I’ve changed every function to the async ab() {} syntax, except for uploadFile, which remains an arrow function to preserve its this binding for the OnlyOffice callback.

apps/api/src/filesharing/filesharingQueueProcessor.ts
```diff
@@ -0,0 +1,106 @@
...
+      const percent = Math.round((processed / total) * 100);
+      const studentName = FilesharingService.getStudentNameFromPath(destinationPath) || '';
+
+      SseService.sendEventToUser(
+        username,
+        this.fileSharingSseConnections,
+        new FilesharingProgressDto(Number(job.id), processed, total, studentName, percent, originFilePath, failedPaths),
+        SSE_MESSAGE_TYPE.UPDATED,
+      );
+    }
```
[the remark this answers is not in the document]
Maddi02: Now each file share is its own job in the queue. 


https://github.com/user-attachments/assets/0a2bfb2f-5060-45d8-957c-f8b5cee562f6

michidold (approved):
Looks good.

apps/api/src/filesharing/filesharingQueueProcessor.ts
TomlDev, line 1: > Consumer vs. Processor
> 
>     Consumer is a common term in pub/sub or messaging systems (e.g., Kafka or RabbitMQ).
>     Processor is a bit more Bull-specific. It suggests that the class/function is responsible for processing the job’s logic rather than simply consuming messages in a generic sense.

I would rename the `FilesharingQueueProcessor` to `FilesharingConsumer` to use the naming that is widely used for this technology. I would not rely on this specific library that we use. It makes it clearer for outside developers.

(and the file to filesharing.consumer.ts because of our naming convention)

apps/api/src/filesharing/filesharingQueueProcessor.ts
```diff
@@ -0,0 +1,107 @@
...
+      const collectFolderExists = await this.fileSharingService.checkIfFileOrFolderExists(
+        username,
+        pathUpToTeacherFolder,
+        FILE_PATHS.COLLECT,
+        ContentType.DIRECTORY,
+      );
+      if (!collectFolderExists) {
+        await this.fileSharingService.createFolder(username, pathUpToTeacherFolder, FILE_PATHS.COLLECT);
+      }
+    }
```
TomlDev, line 82: This should be part of the file sharing service

```
// filesharing.service.ts
async ensureFolderExists(username: string, basePath: string, folderName: string) {
  const exists = await this.checkIfFileOrFolderExists(
    username,
    basePath,
    folderName,
    ContentType.DIRECTORY,
  );
  if (!exists) {
    await this.createFolder(username, basePath, folderName);
  }
}
```

Then here:
```
await this.fileSharingService.ensureFolderExists(username, pathUpToTransferFolder, username);
await this.fileSharingService.ensureFolderExists(username, pathUpToTeacherFolder, FILE_PATHS.COLLECT);

File: apps/api/src/filesharing/filesharingQueueProcessor.ts [hunk 40]
```diff
@@ -0,0 +1,107 @@
...
+    super();
+  }
+
+  subscribe(username: string, res: Response): Observable<MessageEvent> {
+    return SseService.subscribe(username, this.fileSharingSseConnections, res);
+  }
+
+  async process(job: Job<DuplicateFileJobData>): Promise<void> {
+    switch (job.name) {
+      case QUEUE_NAMES.DUPLICATE_FILE_QUEUE:
```
TomlDev, line 45: Is this a queue name or a job name?

apps/frontend/src/components/ui/Table/ScrollableTable.tsx
```diff
@@ -71,6 +76,11 @@ const ScrollableTable = <TData, TValue>({
   enableRowSelection,
   initialSorting,
   tableIsUsedOnAppConfigPage = false,
+  textColorClass = 'text-muted-foreground',
```
TomlDev, line 79: Classname

apps/frontend/src/components/ui/Table/ScrollableTable.tsx
```diff
@@ -223,6 +235,7 @@ const ScrollableTable = <TData, TValue>({
           </Table>
         </div>
       </div>
+      {footer && <div className="max-w-[42vh] overflow-hidden text-ellipsis whitespace-nowrap">{footer}</div>}
```
TomlDev, line 238: Why the max width? What is with wider dialogs? Can you share a screenshot of this in use?

apps/frontend/src/locales/de/translation.json
```diff
@@ -457,7 +467,22 @@
...
     "sharedMailBox": "Geteiltes Postfach",
-    "enrolPageDescription": "Wählen Sie die Klassen und Projekte aus, für die Sie sich einschreiben möchten, indem Sie die Checkboxen aktivieren.",
+    "enrolPageDescription": "Wähle die Klassen und Projekte aus, für die du dich einschreiben möchtest, indem du die Checkboxen aktivierst.",
+
+    "copyOrCut": "Möchtest du die Dateien ausschneiden oder kopieren?",
+    "filesShared": "Dateien geteilt",
+    "filesSharingStarted": "Dateien werden geteilt",
+    "filesCollected": "Dateien eingesammelt",
+    "failDialog": {
+      "title": "Datei „{{file}}“ konnte nicht geteilt werden",
```
TomlDev, line 477: Its the only place where we use the ticks around `„{{file}}“`

apps/frontend/src/locales/de/translation.json
```diff
@@ -457,7 +467,22 @@
...
+      "reasonMissing": "Die Datei ist möglicherweise nicht mehr vorhanden oder wurde umbenannt.",
+      "reasonAlreadyReceived": "Der Empfänger hat die Datei bereits.",
+      "reasonDuplicate": "Der Dateiname ist doppelt vorhanden.",
+      "affectedPersons": "Die folgenden Personen haben die Datei nicht erhalten:",
+      "affectedPerson": "Die folgende Person hat die Datei nicht erhalten:",
+      "filenameAdvice": "Wenn das Austeilen nochmals probiert werden soll, wird die Datei unter dem Namen {{filename}} verteilt",
```
TomlDev, line 483: "erneut versucht" statt nochmals probiert

apps/frontend/src/locales/en/translation.json
```diff
@@ -318,19 +319,28 @@
       "WebDavError": "Data server access failed",
       "DownloadFailed": "Download failed",
       "FolderCreationFailed": "Folder creation failed",
+      "CreateCollectFolderForStudentFailed": "Create collect folder for student failed",
```
TomlDev, line 322: Creating the collect folder ...

apps/frontend/src/locales/en/translation.json
```diff
@@ -318,19 +319,28 @@
...
       "FileNameRequired": "File name is required",
       "NewFileNameRequired": "New file name is required",
       "NameRequired": "Name is required"
+    },
+    "progressBox": {
+      "info": "Info",
+      "title": "Distributing files",
+      "processedInfo": "{{processed}} out of {{total}} files distributed successfully",
+      "fileInfo": "File {{filename}} is being distributed to {{studentName}}",
+      "errorInfo": "{{failed}} file(s) could not be distributed"
```
TomlDev, line 343: In german its only plural not with `(s)`

apps/frontend/src/locales/en/translation.json
```diff
@@ -458,6 +468,16 @@
     "myRoom": "My room",
     "sharedMailBox": "Shared mailbox",
     "enrolPageDescription": "Select the classes and projects that you want to enrol yourself by checking the checkboxes.",
+    "failDialog": {
+      "title": "The file \"{{file}}\" could not be shared",
```
TomlDev, line 472: Here you use different ticks then in german

apps/frontend/src/locales/en/translation.json
```diff
@@ -458,6 +468,16 @@
...
+      "reasonMissing": "The file may no longer exist or has been renamed.",
+      "reasonAlreadyReceived": "The recipient already has the file.",
+      "reasonDuplicate": "The file name is duplicated.",
+      "affectedPersons": "The following persons did not receive the file:",
+      "affectedPerson": "The following person did not receive the file:",
+      "filenameAdvice": "If you want to try sharing again, the file will be shared under the name {{filename}}",
```
TomlDev, line 478: If you try to share again,

apps/frontend/src/locales/en/translation.json
```diff
@@ -458,6 +468,16 @@
...
+      "retryButton": "Try Again"
```
TomlDev, line 479: Retry

apps/frontend/src/locales/en/translation.json
```diff
@@ -468,7 +488,11 @@
       "proxyAdressPlaceholder": "http://localhost:11080",
       "createConfig": "Create config",
       "editConfig": "Edit config"
-    }
+    },
+    "copyOrCut": "Would you like to cut or copy the files?",
+    "filesShared": "Files shared",
+    "filesSharingStarted": "Files sharing started",
```
TomlDev, line 494: This seems wrong
Sharing files started?

apps/frontend/src/pages/ClassManagement/LessonPage/LessonPage.tsx
```diff
@@ -56,12 +63,47 @@ const LessonPage = () => {
...
   const [isPageLoading, setIsPageLoading] = useState(false);
   const [currentSelectedSession, setCurrentSelectedSession] = useState<LmnApiSession | null>(null);
 
+  const [isFileSharingProgessInfoDialogOpen, setIsFileSharingProgessInfoDialogOpen] = useState(false);
+
+  useEffect(() => {
+    if (!filesharingProgress) return;
+
+    const percent = filesharingProgress.percent ?? 0;
+    const dataForBox = {
```
TomlDev, line 79: toasterData?

apps/frontend/src/pages/ClassManagement/LessonPage/LessonPage.tsx
```diff
@@ -120,6 +162,13 @@ const LessonPage = () => {
     }
   }, [groupTypeParams, groupNameParams, lmnApiToken]);
 
+  useEffect(() => {
+    if (currentSelectedSession) {
+      setMember(currentSelectedSession.members);
+    }
+    if (!isLoading) setIsPageLoading(false);
+  }, [userSessions]);
+
```
TomlDev, line 171: I think this is a relict of merging and should be removed

apps/frontend/src/pages/ClassManagement/LessonPage/LessonPage.tsx
```diff
@@ -203,8 +265,26 @@ const LessonPage = () => {
...
+          })}
+          body={
+            <SharingFilesFailedDialogBody
+              failedFilePath={filesharingProgress?.currentFile}
+              affectedUsers={filesharingProgress?.failedPaths.map((path) => path.split('/').at(2) || '')}
+              failedPaths={filesharingProgress?.failedPaths}
+            />
+          }
+        />
+      )}
```
TomlDev, line 287: Could you share a screenshot of this?

apps/frontend/src/pages/FileSharing/Table/FileSharingTableColumns.tsx
```diff
@@ -29,178 +29,195 @@ import ContentType from '@libs/filesharing/types/contentType';
...
+      accessorFn: (row) => row.type + row.filename,
+      cell: ({ row }) => {
+        const [searchParams, setSearchParams] = useSearchParams();
+        const { currentlyDisabledFiles, setFileIsCurrentlyDisabled } = useFileSharingStore();
+        const { resetCurrentlyEditingFile, setPublicDownloadLink, setIsFilePreviewVisible, isFilePreviewDocked } =
+          useFileEditorStore();
+        const isCurrentlyDisabled = currentlyDisabledFiles[row.original.basename];
+        const handleFilenameClick = () => {
+          if (onFilenameClick) {
+            onFilenameClick(row.original);
```
TomlDev, line 87: Better pass only `row` so you can `row.toggleSelected` when it's a file that is clicked. New task:
https://github.com/edulution-io/edulution-ui/issues/626

apps/frontend/src/pages/FileSharing/useFileSharingStore.ts
```diff
@@ -128,6 +130,52 @@ const useFileSharingStore = create<UseFileSharingStore>(
...
+          }
+          return '';
+        } catch (error) {
+          handleApiError(error, set);
+          return '';
+        } finally {
+          set({ isLoading: false });
+        }
+      },
+
```
TomlDev, line 178: This was moved to `apps/frontend/src/pages/FileSharing/FilePreview/OnlyOffice/useFileEditorStore.ts`
Please remove it here.

TomlDev (changes requested):
See my comments

apps/api/src/filesharing/filesharingQueueProcessor.ts
Maddi02: job name

Maddi02: I will rename it

apps/api/src/filesharing/filesharing.service.ts
```diff
@@ -259,42 +279,109 @@ class FilesharingService {
...
+    name: string,
+    contentType: ContentType,
+  ): Promise<boolean> {
+    if (contentType === ContentType.DIRECTORY) {
+      const directories = await this.getDirAtPath(username, `${parentPath}/`);
+      return directories.some((item) => item.type === ContentType.DIRECTORY && item.basename === name);
+    }
+
+    return false;
+  }
```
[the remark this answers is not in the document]
Maddi02: Renamed it to `checkIfFolderExists`

apps/frontend/src/components/ui/Table/ScrollableTable.tsx
```diff
@@ -71,6 +76,11 @@ const ScrollableTable = <TData, TValue>({
   enableRowSelection,
   initialSorting,
   tableIsUsedOnAppConfigPage = false,
+  textColorClass = 'text-muted-foreground',
+  showHeader = true,
+  showSelectedCount = true,
+  footer,
```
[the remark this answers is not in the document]
Maddi02: `MoveContentDialogBody.tsx`

apps/frontend/src/pages/ClassManagement/LessonPage/LessonPage.tsx
Maddi02: image


image

libs/src/filesharing/types/filesharingProgressDto.ts
```diff
@@ -0,0 +1,47 @@
...
+ * This program is free software: you can redistribute it and/or modify it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+
+class FilesharingProgressDto {
```
TomlDev, line 13: As discussed please add the annotations and remove the contructor

apps/frontend/src/pages/ClassManagement/LessonPage/LessonPage.tsx
TomlDev: I meant the SharingFilesFailedDialogBody

libs/src/filesharing/types/filesharingProgressDto.ts
```diff
@@ -10,38 +10,31 @@
  * You should have received a copy of the GNU Affero General Public License along with this program. If not, see <https://www.gnu.org/licenses/>.
  */
 
-class FilesharingProgressDto {
+import { IsNumber, IsString, IsOptional, IsArray } from 'class-validator';
+
+export class FilesharingProgressDto {
```
TomlDev, line 15: remove export

TomlDev (changes requested):
Please have a look 👍🏽

apps/api/src/filesharing/filesharing.consumer.ts
```diff
@@ -70,15 +70,15 @@ class FilesharingConsumer extends WorkerHost {
...
-    const progressDto = new FilesharingProgressDto(
-      Number(job.id),
+    const progressDto: FilesharingProgressDto = {
+      processID: Number(job.id),
       processed,
       total,
-      studentName,
       percent,
-      originFilePath,
+      currentFile: originFilePath,
```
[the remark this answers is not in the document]
Maddi02: Yes. `currentFilePath' is clearer.

apps/frontend/src/pages/ClassManagement/LessonPage/LessonPage.tsx
Maddi02: image

apps/frontend/src/components/ui/Table/ScrollableTable.tsx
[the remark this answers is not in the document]
Maddi02:
```
<ScrollableTable
              ....
              showHeader={false}
              textColorClassname="text-background"
              showSelectedCount={false}
              filterKey="select-filename"
              .....
               />
```

apps/frontend/src/pages/ClassManagement/LessonPage/LessonPage.tsx
```diff
@@ -206,16 +206,9 @@ const LessonPage = () => {
...
-      filesharingProgress.failedPaths?.length &&
-      filesharingProgress.failedPaths?.length > 0
-    ) {
-      setIsFileSharingProgessInfoDialogOpen(true);
-    } else {
-      setIsFileSharingProgessInfoDialogOpen(false);
-    }
+    setIsFileSharingProgessInfoDialogOpen(
+      (filesharingProgress?.percent ?? 0) >= 100 && (filesharingProgress?.failedPaths?.length ?? 0) > 0,
+    );
```
TomlDev, line 211: What about something like this, FFTR
```
const hasProgressCompleted = (filesharingProgress?.percent ?? 0) >= 100;
const hasFailedPaths = (filesharingProgress?.failedPaths?.length ?? 0) > 0;

setIsFileSharingProgessInfoDialogOpen(hasProgressCompleted && hasFailedPaths);

TomlDev (approved):

PR #241, opened 2024-10-30
Implement tests for groups and lmnApi controllers/services

TomlDev:
There were no tests before.


apps/api/src/groups/groups.controller.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+          provide: GroupsService,
+          useValue: mockGroupsService,
+        },
+      ],
+    }).compile();
+
+    controller = module.get<GroupsController>(GroupsController);
+    service = module.get<GroupsService>(GroupsService);
+
+    // Mock the static method
```
michidold, line 25: Comment could be removed

apps/api/src/groups/groups.service.spec.ts
```diff
@@ -0,0 +1,256 @@
+/* eslint-disable @typescript-eslint/dot-notation */
```
michidold, line 1: Why is dot notation not used?
`GroupsService['fetchAllGroups']` -> `GroupsService.fetchAllGroups`

apps/api/src/lmnApi/lmnApi.controller.spec.ts
michidold, line 1: Just two thoughts:
In this controller test, or in any controller test, we simply test, if the function won't throw any error if called correctly. The result of that call is then tested separately in the service.spec. Doesn't it make sense, to only test it in controller.spec?

#FFTR

michidold (approved):
Just small findings and a question.

apps/api/src/groups/groups.service.spec.ts
TomlDev: It allows accessing private static methods. With dot notation you get this:
> TS2341: Property `sanitizeGroup` is private and only accessible within class `GroupsService`

For some methods (like `fetchAllGroups`) it was not necessary, so I changed it back like you wrote. 👍🏽

apps/api/src/lmnApi/lmnApi.controller.spec.ts
TomlDev: Ah you mean that in the service test it shouldn't be necessary to deal with mocked requests since the tests for the service should only test the service and not the controller?
True.

PR #265, opened 2024-11-14
Replace ControlPanel with a Window frame

TomlDev:
See the issue for the previous implementation. It just didn't look very professional/state of the art.

Change to a normal window bar that allows maximizing, minimizing and windowed mode (and closing).

https://github.com/user-attachments/assets/b1b6a1ce-48b1-4ea7-9c69-0e2dc8e71280

Currently conferencing and desktop-deployment are supported (vdi could not be tested because it is not running locally).
And in the file manager the fullscreen preview was also switched:
grafik

### Key changes
- Remove the unused ControlPanel component
- Introduce a new ResizableWindow component
- Add a hook to listen for window size changes
- Rework the frame store to separate embedded frames, that have no windowed mode capability
- Refactor some file manager related files

### Mobile view

https://github.com/user-attachments/assets/a83b8bdc-6875-47e2-844c-59057270b1bc

Todo: https://github.com/edulution-io/edulution-ui/issues/319


apps/frontend/src/components/framing/ResizableWindow/CloseButton.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+const CloseButton = ({ handleClose, className }: { handleClose: () => void; className?: string }) => (
+  <button
+    type="button"
+    onClick={handleClose}
+    className={cn('flex h-10 w-16 items-center justify-center bg-red-800 p-5 hover:bg-red-700', {
+      className,
+    })}
+  >
+    <div className="text-2xl">×</div>
```
michidold, line 12: Have you tested `<MdClose />` from react-icons? Maybe it fits better. Anyways it looks awesome .
##FFTR

apps/frontend/src/pages/DesktopDeployment/DesktopDeploymentPage.tsx
```diff
@@ -85,29 +87,32 @@ const DesktopDeploymentPage: React.FC = () => {
...
-              availableClients={getAvailableClients(os, virtualMachines)}
-              onClick={() => handleConnect()}
-              osType={os}
-              disabled={getAvailableClients(os, virtualMachines) === 0}
-            />
-          ))}
-        </div>
-        <ConnectionErrorDialog handleReload={handleReload} />
-        <LoadingIndicator isOpen={isLoading} />
+      <NativeAppHeader
```
michidold, line 94: A scrollbar appears at least on chrome:
grafik

apps/frontend/src/pages/FileSharing/previews/FileViewer.tsx
```diff
@@ -1,45 +1,43 @@
...
-          showEditor={showEditor}
-          mode={mode}
-          isMobile={isMobile}
-          currentlyEditingFile={currentlyEditingFile}
-        />
+      editMode={isOpenedInNewTab}
+    >
+      {isFullScreenEditingEnabled && !isOpenedInNewTab ? (
+        <ResizableWindow
+          disableMinimizeWindow
```
michidold, line 30: Can we make this minimizeable as well?

libs/src/filesharing/constants/routes.ts
```diff
@@ -0,0 +1,2 @@
+const ONLY_OFFICE_ROUTE = '/onlyOffice';
```
michidold, line 1: I would use lower case and the slash isn't needed i think. But have a look.
##FFTR

apps/frontend/src/components/framing/ResizableWindow/ResizableWindow.tsx
```diff
@@ -0,0 +1,215 @@
...
+  const [isMaximized, setIsMaximized] = useState(true);
+  const [currentPosition, setCurrentPosition] = useState({ x: 0, y: 0 });
+  const [prevPosition, setPrevPosition] = useState(currentPosition);
+
+  const isMinimized = minimizedWindowedFrames.includes(titleTranslationId);
+
+  const [prevSize, setPrevSize] = useState(currentWindowedFrameSizes[titleTranslationId]);
+
+  const setMaxWidth = () => {
+    const width = isMobileView ? window.innerWidth : window.innerWidth - SIDEBAR_WIDTH;
```
michidold, line 53: On smartphone at least VDI is not useable. Think we should block it on mobile phone, but not for tablets

apps/frontend/src/components/framing/ResizableWindow/ResizableWindow.tsx
```diff
@@ -0,0 +1,215 @@
...
+            />
+          )}
+          <CloseButton
+            handleClose={() => {
+              handleClose();
+              setWindowedFrameMinimized(titleTranslationId, false);
+              setWindowedFrameOpen(titleTranslationId, false);
+            }}
+            className={isMinimized ? 'h-5 w-8 text-sm' : ''}
+          />
```
michidold, line 204: On mobile these buttons are not clickable if its minimized.
[Bildschirmaufzeichnung vom 2024-12-20 16-50-53.webm](https://github.com/user-attachments/assets/9fb0bf4d-99e1-47ae-ae0b-c1b35674771b)

michidold (changes requested):
Found some stuff, please have a look.

apps/frontend/src/pages/FileSharing/previews/FileViewer.tsx
TomlDev: The behavior is taken from the previous implementation, I think if we wanted to make it minimizable we have to take more edge cases into consideration. Therefore I would not want to add this change in this PR.

libs/src/filesharing/constants/routes.ts
TomlDev: I replaced the original occurrences 
grafik
grafik

I would not want to introduce any changes in this PR for the lower case. You want me to move the slash outside of the const?

apps/frontend/src/components/framing/ResizableWindow/ResizableWindow.tsx
TomlDev: I think it's a problem of touch emulation:

https://github.com/user-attachments/assets/0430fd69-d660-4a8f-a21f-142d69ad750a

In firefox with touch enabled I had to click twice. Will test it on mobile.

libs/src/filesharing/constants/routes.ts
TomlDev: The navigation still works, I could not find other places where it was used so I hope the change has no other impacts 👍🏽

apps/frontend/src/components/framing/ResizableWindow/ResizableWindow.tsx
TomlDev: With the rework at least the displaying/controls should work. If it makes sense is another question :D

apps/frontend/src/components/framing/ResizableWindow/ResizableWindow.tsx
TomlDev: There was actually a problem with it but it's fixed now. Should work on mobile now correctly.

michidold (approved):
Looks good

PR #271, opened 2024-11-19
Add env to vite config

michidold:
Removed urls from repo and simplified adjustment for other environment

Add [new .env files](https://doku.netzint.de/doku.php?id=intern:technik:schulserver:edulution:guides:dev-env)


TomlDev (approved):
👍🏽

PR #272, opened 2024-11-20
Fix language detection on chrome, fixed missing translations

michidold:
- Fixed the fallback language, chrome uses the html lang="en", which is defined in the index.html.
- Add some missing translations


Maddi02 (approved):
👍

PR #283, opened 2024-11-28
270 Add base for database migrations

maierdominik:
CHANGES:

```
270 Add base for db migrations:
     * add 'migrate' plugin
     * add a script to handle/use the migration cli
            * can be used like: `npm run migrate appconfig create new-migration`, 
              or `npm run migrate appconfig up`
     * create an example for the migration and adjust the structure such that 
       appconfigs, `1732619590704-add-db-version-number`):
            * each module/collection should have it's own migrations called from a new migration service that is 
              linked to the respective controller (here: `appconfig.migration.service.ts`).
            * the service should create a backup before the migration starts (for the rollback)
                   * create a copy of the collection in the same database with the aggregation pipeline 
                      (The user does not have to install mongodump/mongorestore) 
```


maierdominik:
WE DECIDED NOT TO USE ANY PLUGIN

PR #290, opened 2024-12-03
Optimize frontend bundle size

michidold:
- Add manual chunks for heavy libraries
- lazy load Whiteboard and SurveyPage
- Fix issue due to new modules


TomlDev (approved):
👌🏽

PR #297, opened 2024-12-03
270 Add Base for the databse migrations

maierdominik:
New start without using an external plugin

270: 
* adapt schema sucht that each document has the property `schemaVersion`
* create migration where the value for `schemaVersion` is set to `0` to test it


apps/frontend/src/pages/ClassManagement/components/GroupDialog/GroupPropertiesTable.tsx
```diff
@@ -89,7 +89,7 @@ const GroupPropertiesTable = ({ isCreateMode, disabled, form }: GroupPropertiesT
       case 'text':
       default:
         if (groupProperty.disabled) {
-          return <>{watch(groupProperty.name)}</>;
+          return watch(groupProperty.name) as string;
```
TomlDev, line 92: Instead change line 14 to 
`name: keyof Omit<GroupForm, 'admins' | 'admingroups' | 'members' | 'membergroups'>`

apps/api/src/appconfig/appconfig.migration.service.ts
```diff
@@ -0,0 +1,62 @@
...
+    @InjectConnection() private readonly connection: Connection,
+    @InjectModel(AppConfig.name) private readonly appConfigModel: Model<AppConfig>,
+  ) {}
+
+  async onModuleInit() {
+    await initializeCollection(this.connection, this.appConfigModel);
+
+    Logger.log(`⬆ ️⬆️ Migrating`, AppConfigMigrationService.name);
+
+    Logger.log('⬆ ... executing Migrations', AppConfigMigrationService.name);
```
TomlDev, line 19: I would combine the two logging lines
```
    Logger.log(`⬆ ️⬆️ Executing migratings`, AppConfigMigrationService.name);

File: apps/api/src/appconfig/appconfig.migration.service.ts [hunk 3]
```diff
@@ -0,0 +1,62 @@
...
+
+    // eslint-disable-next-line no-underscore-dangle
+    const ids = unprocessedDocuments.map((d) => d._id);
+
+    try {
+      await this.appConfigModel.updateMany(
+        { _id: { $in: ids } },
+        {
+          $set: {
+            migrationNumber: newMigrationNumber,
```
TomlDev, line 49: I dislike the name `migrationNumber`, what about `schemaVersion`, this is also used in the MongoDB docs

apps/api/src/appconfig/appconfig.migration.service.ts
```diff
@@ -0,0 +1,62 @@
...
+            updatedAt: new Date().toDateString(),
+          },
+        },
+      );
+      Logger.log(`⬆ Migration "${migrationName}" was successfully completed.`, AppConfigMigrationService.name);
+    } catch (e) {
+      Logger.error(`⬆     Error while running Migration "${migrationName}": ${e}`, AppConfigMigrationService.name);
```
TomlDev, line 56: Is there a reason for the whitespaces?

TomlDev (changes requested):
Looks good 👍🏽

apps/api/src/appconfig/appconfig.migration.service.ts
maierdominik: Reverted.
My thought were that it is easier to differ this line of output from all the other lines.

TomlDev (approved):
👍🏽

Issue #303, opened 2024-12-09
[BUG] Survey editor not working


michidold:
- Survey editor toggles after opening
- Icons are much too big

[Bildschirmaufzeichnung vom 2024-12-09 14-28-57.webm](https://github.com/user-attachments/assets/4bea0228-5960-4608-8156-91da88c59b18)

PR #306, opened 2024-12-10
303 [BUG] Survey editor not working

maierdominik:
Issue: https://github.com/edulution-io/edulution-ui/issues/303

303 fix survey editor
* fix surveyEditor toggles after opening (-> fix the conversation of the survey formula type)
    * isSurveyFormula does now return true if the title is set and pages and or elements are not defined
    * add a default title to the survey formulas
* add css styles that control the size of the icons

BEFORE:
20241210_Fix-survey-editor_BEFORE

AFTER:
20241210_Fix-survey-editor_AFTER


michidold (approved):
Looks good

PR #310, opened 2024-12-11
Fix loading of appconfig

TomlDev:


apps/frontend/src/pages/Settings/AppConfig/AppConfigPage.tsx
michidold, line 1: In line 71 the options should be removed as well:
image
Problem is, that inside ProxyConfigForm not the key is used, instead the item itself. You can quick fix that and we rework it in another issue

Issue #246, opened 2024-11-04
246 Classes Section Overflows Instead of Becoming Scrollable - Dashboard


Maddi02:
**Describe the bug**
In the edu UI, when a large number of classes are assigned to a user, the classes section grows in height instead of becoming scrollable. This causes the entire dashboard to become scrollable, rather than limiting the scroll functionality to the classes section itself.

**To Reproduce**
1. Log in to the edu UI with the 'brs-netzint-teacher`
2. Observe that the classes section expands vertically, making the entire dashboard scrollable instead of only the classes section.

**Expected behavior**
The classes section should maintain a fixed height and become scrollable when the number of assigned classes exceeds the visible area. This would allow the dashboard to retain its structure without making the entire interface scrollable.

**Screenshots**
<img width="2408" alt="image" src="https://github.com/user-attachments/assets/92883814-0805-432a-ac90-fd9e1cc11bb4">

PR #314, opened 2024-12-12
246 Classes Section Overflows Instead of Becoming Scrollable - Dashboard

maierdominik:
Issue: https://github.com/edulution-io/edulution-ui/issues/246

246:
* adjust gaps and margins to unify the styling of the dashboard cards
* assign an id to the `Groups` card such that the height is known
* calculate the space that is available for the content of the card and use it


apps/frontend/src/components/shared/Card/card.module.scss
```diff
@@ -15,7 +15,7 @@
     bottom: 0;
     left: 0;
     z-index: -1;
-    margin: -$border;
+    margin: -3px;
```
michidold, line 18: The border should be 4px like on the other cards. And I would keep the variable $border

apps/frontend/src/pages/Dashboard/MobileFileAccess/MobileFileAccessCard.tsx
michidold, line 1: This could be a bit shorter:
grafik

apps/frontend/src/pages/Dashboard/Home.tsx
```diff
@@ -1,21 +1,59 @@
...
+
+  const column1 = (
+    <div className="basis-1/4">
+      <AccountInformation />
+    </div>
+  );
+
+  const column2 = (
+    <div className="basis-1/2">
+      <div className="flex flex-col gap-8">
```
michidold, line 31: These two could be combined

michidold (changes requested):
Seems the cards are not equally high

apps/frontend/src/components/shared/Card/card.module.scss
maierdominik: Reverted,
but in my experience, the border of the other elements seem more narrow than this one with the gradient effect, after the render.

apps/frontend/src/pages/Dashboard/MobileFileAccess/MobileFileAccessCard.tsx
maierdominik: I am not exactly sure how you mean it or what elements you mean,
in hope of meating your expecations i set the minimum height too `200` (previously 280)

image

michidold: Sorry, thought its obviouse -> The mobil access card and the classes card are not the same size

apps/frontend/src/pages/Dashboard/DashboardPage.tsx
```diff
@@ -28,28 +64,10 @@ const DashboardPage: React.FC = () => {
...
-              </div>
-              <div>
-                <QuotaCard />
-              </div>
-            </div>
-          </div>
-          <div className="flex-1">
-            <Feed />
-          </div>
+        <div className={`flex flex-col-reverse gap-8 md:flex-row md:max-h-[${pageBarsHeight}px]`}>
```
michidold, line 67: These two divs could be combined.

apps/frontend/src/pages/Dashboard/DashboardPage.tsx
```diff
@@ -1,21 +1,57 @@
...
 import Groups from './Groups';
 
 const DashboardPage: React.FC = () => {
   const { t } = useTranslation();
+
   const isMobileView = useIsMobileView();
+
   const { user } = useUserStore();
 
+  const pageBarsHeight = useElementHeight([BLANK_LAYOUT_HEADER_ID, DASHBOARD_DESCRIPTION_ID, FOOTER_ID]);
```
michidold, line 21: There is a error thrown by this hook:
[Bildschirmaufzeichnung vom 2024-12-18 09-34-40.webm](https://github.com/user-attachments/assets/8c1ca56d-e168-4aa2-81aa-58f3209ba227)

apps/frontend/src/pages/Dashboard/Groups.tsx
```diff
@@ -1,26 +1,36 @@
...
-            onChange={() => {}}
-            readOnly
-          />
+        <div className="flex flex-col gap-2 p-0">
+          <h4 className="mb-4 font-bold">{t('groups.classes')}</h4>
+          <div
+            className="overflow-y-auto scrollbar-thin"
+            style={{ height: `${cardContentHeight}px` }}
+          >
+            <BadgeField
```
michidold, line 28: Think the error mentioned above occurs, because the badge field is rendered first and then the height is updated, which causes a render loop.

michidold (changes requested):
Had a error during render. Have a look

apps/frontend/src/pages/Dashboard/Groups.tsx
maierdominik: The concerning element is not  the BadgeField but the dashboard description 
 `DASHBOARD_DESCRIPTION_ID`

apps/frontend/src/pages/Dashboard/DashboardPage.tsx
maierdominik: I removed the hook
If the size should adjust itself, I would create a new Issue for this task

apps/frontend/src/components/shared/Card/Card.tsx
```diff
@@ -42,7 +42,7 @@ const CardContent = React.forwardRef<HTMLDivElement, React.HTMLAttributes<HTMLDi
     return (
       <SHCardContent
         ref={ref}
-        className={cn('p-[20px]', className)}
+        className={cn('h-full w-full p-[20px]', className)}
```
michidold, line 45: Think this has no effect?

apps/frontend/src/pages/Dashboard/DashboardPage.tsx
```diff
@@ -1,18 +1,51 @@
...
 import Groups from './Groups';
 
 const DashboardPage: React.FC = () => {
   const { t } = useTranslation();
+
   const isMobileView = useIsMobileView();
+
   const { user } = useUserStore();
 
+  const column1 = (
```
michidold, line 19: This card is still too big

apps/frontend/src/pages/Dashboard/DashboardPage.tsx
```diff
@@ -1,18 +1,51 @@
...
+  );
+
+  const column2 = (
+    <div className="flex basis-1/2 flex-col gap-8">
+      <div className="flex flex-col justify-between gap-8 md:flex-row">
+        <div
+          id={GROUPS_ID}
+          className="flex-1"
+        >
+          <Groups />
```
michidold, line 32: On mobile view there is nothing shown:
grafik

michidold (changes requested):
Found some stuff, please have a look

PR #321, opened 2024-12-16
318 Deleting a bulletin should remove all attachments

TomlDev:
I had to remove/avoid the usage of the bulletin service in the category service because of circular dependencies.
Now the request from the user first tries to remove all bulletins and if this succeeds, he removes the category.


apps/api/src/bulletinboard/bulletinboard.service.ts
```diff
@@ -200,20 +216,45 @@ class BulletinBoardService implements OnModuleInit {
...
-    await this.bulletinModel.deleteMany({ _id: { $in: ids } }).exec();
+    try {
+      await Promise.all(
+        bulletins.map(async (bulletin) => {
+          if (bulletin.attachmentFileNames?.length) {
+            await Promise.all(
+              bulletin.attachmentFileNames.map(async (fileName) => {
+                const filePath = join(this.attachmentsPath, fileName);
+                if (existsSync(filePath)) {
+                  await unlink(filePath);
```
michidold, line 243: I would prefer using the promises.unlink from "fs", since it is a natve node function and not a library. 
`import {promises} from "fs"`
`await promises.unlink(filePath);`

michidold (approved):
Only one small finding

PR #322, opened 2024-12-16
320 Clicking on images on bulletin board should open a preview

TomlDev:


apps/frontend/src/components/shared/ImageModal.tsx
```diff
@@ -0,0 +1,42 @@
...
+}
+
+/*
+ TODO: Change to window frame after #263 is merged, see https://github.com/edulution-io/edulution-ui/issues/319
+ */
+const ImageModal: React.FC<ImageModalProps> = ({ isOpen, imageUrl, onClose }) => {
+  if (!isOpen) return null;
+
+  return createPortal(
+    // eslint-disable-next-line jsx-a11y/click-events-have-key-events,jsx-a11y/no-static-element-interactions
```
michidold, line 17: Is it possible to use `useOnClickOutside` from useTsHooks? To avoid these click listeners on the div elements?

michidold (approved):
Looks good, just one question

apps/frontend/src/components/shared/ImageModal.tsx
TomlDev: I hope to replace it fast with our new component when it's merged so I think it's okay to leave it until then, what do you think?

TomlDev: Resolved as discussed

PR #324, opened 2024-12-16
Fix-mail-sync-not-configured

michidold:


apps/api/src/mails/mails.service.ts
```diff
@@ -196,14 +196,18 @@ class MailsService {
   }
 
   async getSyncJobs(username: string) {
+    if (!MAIL_API_URL || !MAIL_API_KEY) {
+      return [] as SyncJobDto[];
```
TomlDev, line 200: The cast is not necessary. Better define a return type of the function to make sure it's always that:
```
  async getSyncJobs(username: string): Promise<SyncJobDto[]> {

TomlDev (approved):

PR #329, opened 2024-12-17
Small refactoring of logging

TomlDev:
- Replaces some lines where the catched `error` was passed in the `data` object of the custom error.


michidold (approved):
Looks good. Think we should also agree to always use the same error alias -> not e, err, error

Issue #274, opened 2024-11-22
Dashboard - Mail-Feed took very long with large mailbox


hermanntoast:
**Describe the bug**
The Dashboard-Mail-Feed took very long to get the unread mails from large mailbox. Maybe it helps to filter the mail for a specified timespan?

TomlDev:
Thanks for the contribution. @maierdominik can you take a look at this?

PR #334, opened 2024-12-19
NIEDUUI-348: IMAP: Retrieve the data for the imap client from the backend

maierdominik:
JIRA: https://schulz-informatik.atlassian.net/browse/NIEDUUI-348

NIEDUUI-348:
    * move imap data from env to db


apps/api/src/mails/mails.service.ts
```diff
@@ -37,20 +39,34 @@ class MailsService {
   }
 
   async getMails(username: string, password: string): Promise<MailDto[]> {
+    const appConfigs = await this.appConfigModel.find();
+
+    const imapOptions = getExtendedOptionValue(
```
maierdominik, line 44: As i understand the function it should return a single value for a extended options field. But it does not.

Does this really work for the other occurances.

libs/src/appconfig/constants/extendedOptions/imapMailFeed.ts
```diff
@@ -0,0 +1,36 @@
...
+    value: '',
+  },
+  {
+    name: ExtendedOptionKeys.MAIL_IMAP_TLS_REJECT_UNAUTHORIZED,
+    title: 'appExtendedOptions.onlyOfficeJwtSecretTitle',
+    description: 'appExtendedOptions.onlyOfficeJwtSecretDescription',
+    type: ExtendedOptionField.password,
+    value: '',
+  },
+];
```
michidold, line 34: Here the title and descriptions are wrong. ExtendedOptionField needs to be type input. Please add a task where we add:
- input type number
- boolean Switch
as options

libs/src/dashboard/constants/pull-time-interval.ts
```diff
@@ -3,4 +3,4 @@ const FEED_PULL_TIME_INTERVAL = 10000;
 
 export default FEED_PULL_TIME_INTERVAL;
 
-export const FEED_PULL_TIME_INTERVAL_SLOW = 60000;
+export const FEED_PULL_TIME_INTERVAL_SLOW = 5000;
```
michidold, line 6: I would keep 60s

michidold (changes requested):
Found some stuff, please have a look

maierdominik:
Another Certification Error:

image

libs/src/appconfig/constants/extendedOptions/imapMailFeed.ts
michidold: Its still password type

apps/frontend/src/locales/de/translation.json
```diff
@@ -1027,6 +1031,7 @@
   },
   "mails": {
     "errors": {
+      "NotAbleToGetImapOption": "Auslesen der IMAP configuration ist nicht möglich",
```
michidold, line 1034: "IMAP-Konfiguration"

apps/api/src/mails/mails.service.ts
```diff
@@ -36,21 +47,36 @@ class MailsService {
...
+  @OnEvent(EVENT_EMITTER_EVENTS.APPCONFIG_UPDATED)
+  async updateImapConfig() {
+    try {
+      const appConfigs = await this.appConfigService.getAppConfigs([GroupRoles.SUPER_ADMIN]);
+      this.imapUrl = getExtendedOptionValue(appConfigs, APPS.MAIL, ExtendedOptionKeys.MAIL_IMAP_URL);
+      this.imapPort = Number(getExtendedOptionValue(appConfigs, APPS.MAIL, ExtendedOptionKeys.MAIL_IMAP_PORT));
+      this.imapSecure = Boolean(getExtendedOptionValue(appConfigs, APPS.MAIL, ExtendedOptionKeys.MAIL_IMAP_SECURE));
+      this.imapRejectUnauthorized = Boolean(
+        getExtendedOptionValue(appConfigs, APPS.MAIL, ExtendedOptionKeys.MAIL_IMAP_TLS_REJECT_UNAUTHORIZED),
+      );
```
michidold, line 63: This returns true even if its false in db. Change it to:
```typescript
this.imapRejectUnauthorized =
        getExtendedOptionValue(appConfigs, APPS.MAIL, ExtendedOptionKeys.MAIL_IMAP_TLS_REJECT_UNAUTHORIZED) === 'true';
```

apps/api/src/mails/mails.service.ts
```diff
@@ -36,21 +47,36 @@ class MailsService {
...
+    } catch (error) {
+      throw new CustomHttpException(MailsErrorMessages.NotAbleToGetImapOption, HttpStatus.INTERNAL_SERVER_ERROR);
+    }
+  }
+
   async getMails(username: string, password: string): Promise<MailDto[]> {
-    // TODO: NIEDUUI-348: Migrate this settings to AppConfigPage (set imap settings in mails app config)
-    if (!MAIL_IMAP_URL || !MAIL_IMAP_PORT || !MAIL_IMAP_SECURE || !MAIL_IMAP_TLS_REJECT_UNAUTHORIZED) {
+    if (!this.imapUrl || !this.imapPort || !this.imapSecure || !this.imapRejectUnauthorized) {
```
michidold, line 70: if its parsed to boolean the imapSecure and imapReject.. will throw a true, so we should remove them.
`if (!this.imapUrl || !this.imapPort)`

apps/api/src/mails/mails.service.ts
```diff
@@ -80,16 +106,20 @@ class MailsService {
...
       const fetchMail: AsyncGenerator<FetchMessageObject> = this.imapClient.fetch(
-        { or: [{ new: true }, { seen: false }, { recent: true }] },
-        {
-          source: true,
-          envelope: true,
-          bodyStructure: true,
-          flags: true,
-          labels: true,
-        },
+        { seen: false, since: new Date(Date.now() - 7 * 24 * 60 * 60 * 1000) },
```
michidold, line 109: I would not filter for the last 7 days here. Maybe we set a filter in the frontend in future. Do we show all mails in mailsfeed?

apps/frontend/src/locales/de/translation.json
```diff
@@ -1048,7 +1053,15 @@
     "onlyOfficeUrl": "Bitte gib die OnlyOffice-URL an",
     "onlyOfficeUrlTitle": "OnlyOffice-URL",
     "onlyOfficeJwtSecretTitle": "OnlyOffice JWT Secret",
-    "onlyOfficeJwtSecretDescription": "Bitte gib den OnlyOffice JWT Secret Key an"
+    "onlyOfficeJwtSecretDescription": "Bitte gib den OnlyOffice JWT Secret Key an",
+    "mailImapUrlTitle": "URL",
+    "mailImapUrlDescription": "Bitte gib eine gültige URL an um die Mailbox per IMAP zu verbinden",
```
michidold, line 1058: Add the fact that http/https should be avoided, only the FQDN is needed.

michidold (changes requested):
Some findings left, please test your implementation before requesting a review.

libs/src/appconfig/constants/extendedOptions/imapMailFeed.ts
maierdominik: The new Input types are added in following PR:
https://github.com/edulution-io/edulution-ui/pull/344

apps/api/src/mails/mails.service.ts
maierdominik: All that are in the inbox and unseen, therefore the `seen` flag.
I added the seven days limit because of the following issue:
https://github.com/edulution-io/edulution-ui/issues/274

I will create a new branch with this filter in order to test it.

apps/frontend/src/locales/de/translation.json
maierdominik: like this ?
image

michidold: The description is not displayed in AppConfig Page.

michidold (changes requested):
One last thing, the description is missing in the appconfig page

michidold (approved):
Looks good

PR #336, opened 2024-12-19
Add thumbnail to avatar

michidold:
- Added config for profile picture
- Added thumbnail to Avatar component
- Added avatar to UserCard in Class Management
- Reworked patch user details query


apps/frontend/src/pages/ClassManagement/LessonPage/UserArea/UserCard.tsx
```diff
@@ -92,8 +93,11 @@ const UserCard = ({
             )}
             onClick={(event) => event.stopPropagation()}
           >
-            {givenName.slice(0, 1)}
-            {surname.slice(0, 1)}
+            <Avatar
+              user={{ username: name, firstName: givenName, lastName: surname }}
+              imageSrc={thumbnailPhoto}
+              className={thumbnailPhoto && 'h-24 w-24 p-2'}
+            />
```
TomlDev, line 100: When I try to upload an image I get a 404 error:
grafik

apps/frontend/src/pages/UserSettings/Details/UserImageConfig.tsx
```diff
@@ -0,0 +1,80 @@
...
+              {t('common.delete')}
+            </Button>
+          ) : null}
+          <Button
+            variant="btn-security"
+            size="lg"
+            onClick={handleSubmitImage}
+            disabled={!base64Image}
+          >
+            {t('common.save')}
```
TomlDev, line 72: In general: Shouldn't it save the image in the moment of selecting? 
Currently the user has to press save. Does that make sense? 
Or has it any advantages compared to upload the image just in the moment where it is selected?

apps/frontend/src/pages/UserSettings/Details/UserImageConfig.tsx
```diff
@@ -0,0 +1,80 @@
...
+import getCompressedImage from '@/utils/getCompressedImage';
+import Avatar from '@/components/shared/Avatar';
+import { toast } from 'sonner';
+
+const UserImageConfig: React.FC = () => {
+  const { t } = useTranslation();
+  const { user, patchUserDetails } = useLmnApiStore();
+  const [base64Image, setBase64Image] = useState<string>(
+    user?.thumbnailPhoto && user?.thumbnailPhoto !== 'undefined' ? user?.thumbnailPhoto : '',
+  );
```
TomlDev, line 14: This could be simplified:
```
const [base64Image, setBase64Image] = useState<string>(user?.thumbnailPhoto ?? '');

File: apps/frontend/src/pages/UserSettings/Details/UserImageConfig.tsx [hunk 4]
```diff
@@ -0,0 +1,80 @@
...
+
+  const handleImageUpload = async (event: React.ChangeEvent<HTMLInputElement>) => {
+    const file = event.target.files?.[0];
+    if (file) {
+      let compressedBase64: string;
+      try {
+        compressedBase64 = await getCompressedImage(file, 100);
+      } catch (error) {
+        toast.error(t(error instanceof Error ? error.message : ''));
```
TomlDev, line 23: I think if you want to also handle a case where error is not an instance of Error you should add a fallback string like 'Image could not be uploaded.' instead of an empty string.

apps/frontend/src/utils/getCompressedImage.ts
```diff
@@ -0,0 +1,69 @@
...
+      img.onload = () => {
+        const canvas = document.createElement('canvas');
+        const ctx = canvas.getContext('2d');
+        if (!ctx) {
+          reject(new Error('usersettings.errors.notSupportedFileFormat'));
+          return;
+        }
+
+        let { width } = img;
+        let { height } = img;
```
TomlDev, line 30: let { width, height } = img;

TomlDev (approved):
👍🏽

apps/frontend/src/pages/UserSettings/Details/UserImageConfig.tsx
michidold: I need that 'undefined' string, because I set that field to undefined if the image is deleted.

PR #339, opened 2024-12-20
211 improve caching when building docker images

michidold:
Unfortunately the github summary shows 0% after all. But build time reduced by around 40 seconds.

Before:
Bildschirmfoto vom 2024-12-20 11-00-29
After:
grafik


TomlDev (approved):
Nice 👍🏽

michidold:
Sorry, I changed only api and frontend tag builds. Will rework the build and test script as well

.github/workflows/build-and-test.yml
```diff
@@ -160,20 +163,21 @@ jobs:
...
+        uses: actions/checkout@v4.2.2
         with:
           token: ${{ github.token }}
 
       - name: Use Node.js
-        uses: actions/setup-node@v4.0.3
+        uses: actions/setup-node@v4.1.0
         with:
-          node-version: 20.x
+          node-version: 18.x
```
TomlDev, line 173: Is there a reason for downgrading node here?

TomlDev (changes requested):
Looks good, only the downgrad of node was not part of the PR description

.github/workflows/build-and-test.yml
michidold: Think we should use the same version for dev and deployment. The release packages are built with 18 as well, so we should do the same for the build and test workflow.

TomlDev (approved):

PR #340, opened 2024-12-20
Add event emitter module

michidold:
- Add module to be used by other branches
- Fixed some logging
- Fixed mobile button not responding in Mail iFrame on mobile


apps/api/src/appconfig/appconfig.guard.ts
```diff
@@ -12,15 +12,15 @@ class AppConfigGuard implements CanActivate {
     const { user } = request;
 
     if (!user) {
-      throw new CustomHttpException(AuthErrorMessages.Unknown, HttpStatus.NOT_FOUND);
+      throw new CustomHttpException(AuthErrorMessages.Unknown, HttpStatus.NOT_FOUND, '', AppConfigGuard.name);
```
TomlDev, line 15: What do you think about passing `undefined` as data or `null`? I find it more clear because a developer knows that the intention is to skip this parameter (`undefined` would be best for this).
With the string I first had to check the method why an empty string would make sense. It's not a big deal, so FFTR

libs/src/common/constants/eventEmitterEvents.ts
TomlDev, line 1: I think every domain should have it's own list of events, what do you think?
So this would be moved to `libs/src/appconfig/constants`

TomlDev (approved):
I could not test the functionality, since `@OnEvent` is not used.

Issue #293, opened 2024-12-03
293 - Check why excalidraw loads from a CDN


TomlDev:
Check the network tab when opening edulution, you see that it downloads a JS file from
https://unpkg.com/@excalidraw/excalidraw@0.17.6/dist/excalidraw-assets-dev/vendor-39727f4653a274cf18f6.js

This should not happen!

maierdominik:
As I understand, it was necessary to copy the excalidraw assests into a specifique folder and update those from time to time.  In the past there was a hard coded path to get there.
They then enabled the download via CDN (unpkg) to simplify the load out of many users. Gladly, they 
always cared about their users that want to self host stuff.

So we could avoid it be setting the `__webpack_public_path_ `

Readme
![image](https://github.com/user-attachments/assets/4631ef49-f829-4bd6-b9f1-2cb0954cf329)

Installation Guide
![image](https://github.com/user-attachments/assets/30c33b5b-9afa-4eb7-818c-189216d6cd25)

Sources:
https://github.com/excalidraw/excalidraw/pull/2835

https://github.com/excalidraw/excalidraw/blob/2af32219740b40633af023b7e8a17cb376f531cf/packages/excalidraw/README.md?plain=1#L23

https://github.com/excalidraw/excalidraw/blob/2af32219740b40633af023b7e8a17cb376f531cf/dev-docs/docs/%40excalidraw/excalidraw/installation.mdx#L27

maierdominik:
Official CodeBox example:
https://codesandbox.io/p/sandbox/excalidraw-ehlz3?file=%2Fpublic%2Findex.html%3A27%2C14-27%2C35

HERE:
The path is set inside of an `index.html` file in the parent folder of the `exaclidraw-assets-dev` directory.
I guess that we would have to copy them from the `node-module`-s folder into a public folder. to pass them similary

PR #343, opened 2025-01-02
293 - Check why excalidraw loads from a CDN

maierdominik:
Issue: https://github.com/edulution-io/edulution-ui/issues/293

293:
    * add exalidraws asset folders to the public folders
    * add the path to the new asset folder to the entry file of the frontend
    * exlude the new asset folder from eslint

BEFORE:
image

AFTER:
image


michidold (changes requested):
First of all, can you remove theses Windows meta files `*:Zone.Identifier`

maierdominik:
Load assets from the node modules folder so that it is always up to date

TomlDev (approved):

michidold (approved):

Issue #342, opened 2024-12-20
342 - Enable a boolean input type with a switch component for the extended option configuration (additionally: half-width)


maierdominik:
Expand `AppConfigExtendedOption` with the types: boolean, number

Add a `Switch component` to set the boolean value.

Add the `width` property to decide if the input field should fill the whole width or only the half width

PR #344, opened 2025-01-02
342 - Enable a boolean input type with a switch component for the extended option configuration

maierdominik:
Issue: https://github.com/edulution-io/edulution-ui/issues/342

342:
    * add an AppConfigFormField for the number and the boolean types
    * add the field width to the option properties
        * use the width inside of the ExtendedOptionsForm.tsx to render half sized input components


apps/frontend/src/components/shared/Input.tsx
```diff
@@ -25,22 +25,33 @@ type InputProps = React.InputHTMLAttributes<HTMLInputElement> &
...
+            },
+          });
+        } else if (type === 'number') {
+          const newValue = Number(value);
+          onChange({
+            ...event,
+            target: {
+              ...event.target,
+              // @ts-expect-error - number type
+              value: newValue,
```
michidold, line 51: I think this transformation is not needed and so also the eslint exception:
```typescript
else if (type === 'number') {
          onChange({
            ...event,
            target: {
              ...event.target,
              value,
            },
          });
```

apps/frontend/src/pages/Settings/AppConfig/components/booleanField/AppConfigSwitch.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { FormControl, FormDescription, FormFieldSH, FormItem, FormMessage } from '@/components/ui/Form';
+import Switch from '@/components/ui/Switch';
+
+type AppConfigSwitchProps<T extends FieldValues> = {
+  fieldPath: Path<T>;
+  control: Control<z.infer<typeof formSchema>, T>;
+  option: AppConfigExtendedOption;
+};
+
+const AppConfigSwitch = <T extends FieldValues>({ fieldPath, control, option }: AppConfigSwitchProps<T>) => {
```
michidold, line 16: Can you add this switch a little padding, so it is in right in the middle of the input field and the description text starts on the same height?
grafik

apps/frontend/src/pages/Settings/AppConfig/components/textField/AppConfigFormField.tsx
```diff
@@ -11,7 +11,7 @@ type AppConfigFormFieldProps<T extends FieldValues> = {
   fieldPath: Path<T>;
   control: Control<z.infer<typeof formSchema>, T>;
   option: AppConfigExtendedOption;
-  type?: 'password' | 'text';
+  type?: HTMLInputTypeAttribute | undefined;
```
michidold, line 14: No need to add `| undefined`, since its optional

libs/src/appconfig/constants/extendedOptionField.ts
```diff
@@ -2,6 +2,8 @@ const ExtendedOptionField = {
   input: 'input',
   password: 'password',
   table: 'table',
+  number: 'number',
+  boolean: 'boolean',
```
michidold, line 6: Maybe we call it "switch"

libs/src/appconfig/types/appConfigExtendedOption.ts
```diff
@@ -1,9 +1,12 @@
...
+import TAppFieldType from '@libs/appconfig/types/tAppFieldType';
+import TAppFieldWidth from '@libs/appconfig/types/tAppFieldWidth';
 
 export interface AppConfigExtendedOption {
   name: string;
   title: string;
   description: string;
   type: ExtendedOptionFieldType;
-  value: string;
+  value: TAppFieldType;
```
michidold, line 10: Make this optional, so you can remove undefined from `TAppFieldType`

libs/src/appconfig/types/tAppFieldType.ts
```diff
@@ -0,0 +1,3 @@
+type TAppFieldType = string | number | boolean | undefined;
```
michidold, line 1: See comment above

libs/src/appconfig/utils/getExtendedOptionValue.ts
```diff
@@ -12,7 +12,7 @@ const getExtendedOptions = (
     return '';
   }
 
-  return appConfig.extendedOptions[key] || '';
+  return String(appConfig.extendedOptions[key]) || '';
```
michidold, line 15: I would not convert it to a string, since it returns "undefined" or "null" strings. If its really null/undefined we want to return a empty string, so it should be casted:

```typescript
  return appConfig.extendedOptions[key] as string || '';

```

michidold (changes requested):
Found some stuff, please have a look.

apps/api/src/mails/mails.service.ts
```diff
@@ -52,13 +51,18 @@ class MailsService implements OnModuleInit {
 
   @OnEvent(EVENT_EMITTER_EVENTS.APPCONFIG_UPDATED)
   async updateImapConfig() {
+    const appConfigs = await this.appConfigService.getAppConfigs([GroupRoles.SUPER_ADMIN]);
```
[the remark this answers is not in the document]
maierdominik: The `appConfigService.getAppConfigs()` has it's own try and catch. 
The surrounding `try- and catch-block` hides the former thrown `AppConfigErrorMessages.ReadAppConfigFailed, HttpStatus.SERVICE_UNAVAILABLE,`.

apps/frontend/src/locales/de/translation.json
```diff
@@ -1060,9 +1060,9 @@
     "mailImapPortTitle": "Port",
     "mailImapPortDescription": "Gib hier die Port Nummer des IMAP-Servers an (z.B. 993)",
     "mailImapSecureTitle": "Sichere Verbindung",
-    "mailImapSecureDescription": "Gib 'true' ein, wenn die Verbindung über TLS aufgebaut werden soll, ansonsten wird 'false' (STARTTLS) verwendet.",
+    "mailImapSecureDescription": "Soll die Verbindung über 'TLS' ('sicher') oder 'STARTTLS' aufgebaut werden.",
     "mailImapRejectUnauthorizedTitle": "Nicht zertifizierte Verbindungen ablehnen",
-    "mailImapRejectUnauthorizedDescription": "Gib 'true' ein, wenn eine Zertifikatsprüfung für die Verbindungen gemacht werden soll. Ansonsten muss hier 'false' gesetzt werden."
+    "mailImapRejectUnauthorizedDescription": "Sollen selbst zertifizierte Verbindungen abgewiesen werden?"
```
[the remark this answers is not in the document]
maierdominik: eng:
image

apps/frontend/src/components/shared/Input.tsx
maierdominik: It is necessary, if we do not want a conversion into a string.

In other words:
If you remove this conversion and save the config, than the value in the database will be of type string and not of type number (Feel free to try it).

apps/frontend/src/pages/Settings/AppConfig/components/booleanField/AppConfigSwitch.tsx
maierdominik: I was not quite sure what you ment with middle:

It now looks like this:
image

maierdominik: Or did you mean like this??:
image

apps/api/src/mails/mails.service.ts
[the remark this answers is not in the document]
michidold: Not directly, the error is passed to the upper catch block which then throws this error. Since the remaining functions won't throw any error, you should remove the try catch block entirely.

apps/frontend/src/locales/de/translation.json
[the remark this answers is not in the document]
michidold: "Should a certificate check be carried out for this connection?"

apps/frontend/src/components/shared/Input.tsx
michidold: Ok, find it also useful to save it as number in the database. Then you can avoid using the eslint exception by casting it:
```typescript
else if (type === 'number') {
          const newValue = Number(value);
          onChange({
            ...event,
            target: {
              ...event.target,
              value: newValue as unknown as string,
            },
          });
        }
        ```

File: apps/api/src/mails/mails.service.ts [hunk 10]
```diff
@@ -52,13 +51,19 @@ class MailsService implements OnModuleInit {
...
-      this.imapPort = Number(getExtendedOptionValue(appConfigs, APPS.MAIL, ExtendedOptionKeys.MAIL_IMAP_PORT));
-      this.imapSecure = getExtendedOptionValue(appConfigs, APPS.MAIL, ExtendedOptionKeys.MAIL_IMAP_SECURE) === 'true';
-      this.imapRejectUnauthorized =
-        getExtendedOptionValue(appConfigs, APPS.MAIL, ExtendedOptionKeys.MAIL_IMAP_TLS_REJECT_UNAUTHORIZED) === 'true';
+      this.imapUrl = String(appConfig.extendedOptions[ExtendedOptionKeys.MAIL_IMAP_URL]);
+      this.imapPort = Number(appConfig.extendedOptions[ExtendedOptionKeys.MAIL_IMAP_PORT]);
+      this.imapSecure = Boolean(appConfig.extendedOptions[ExtendedOptionKeys.MAIL_IMAP_SECURE] || true);
+      this.imapRejectUnauthorized = Boolean(
+        appConfig.extendedOptions[ExtendedOptionKeys.MAIL_IMAP_TLS_REJECT_UNAUTHORIZED] || false,
+      );
```
michidold, line 66: Be careful with conversion, the String() function converts also undefined to "undefined". So all checks with `!this.imapUrl` will be true. Also the Boolean() results in true if there is anything in that field. I would use the following:
```typescript
@OnEvent(EVENT_EMITTER_EVENTS.APPCONFIG_UPDATED)
  async updateImapConfig() {
    const appConfigs = await this.appConfigService.getAppConfigs([GroupRoles.SUPER_ADMIN]);
    const appConfig = appConfigs.find((config) => config.name === APPS.MAIL);
    if (!appConfig || typeof appConfig.extendedOptions !== 'object') {
      throw new CustomHttpException(MailsErrorMessages.NotAbleToGetImapOption, HttpStatus.INTERNAL_SERVER_ERROR);
    }

    this.imapUrl = (appConfig.extendedOptions[ExtendedOptionKeys.MAIL_IMAP_URL] as string) || '';
    this.imapPort = (appConfig.extendedOptions[ExtendedOptionKeys.MAIL_IMAP_PORT] as number) || 0;
    this.imapSecure = appConfig.extendedOptions[ExtendedOptionKeys.MAIL_IMAP_SECURE] === 'true' || false;
    this.imapRejectUnauthorized =
      appConfig.extendedOptions[ExtendedOptionKeys.MAIL_IMAP_TLS_REJECT_UNAUTHORIZED] === 'true' || false;
  }
        ```

File: apps/frontend/src/pages/Settings/AppConfig/components/booleanField/AppConfigSwitch.tsx [hunk 11]
```diff
@@ -0,0 +1,44 @@
...
+
+  return (
+    <FormFieldSH
+      control={control}
+      name={fieldPath}
+      render={({ field }) => (
+        <FormItem>
+          <div>{t(option.title)}</div>
+          <FormControl>
+            <div className="flex h-[36px] items-center">
```
michidold, line 27: use "h-9" instead of "h-[36px]"

apps/frontend/src/pages/Settings/AppConfig/components/ExtendedOptionsForm.tsx
```diff
@@ -71,11 +93,21 @@ const ExtendedOptionsForm = <T extends FieldValues>({
...
-                  {options?.map((option: AppConfigExtendedOption) => renderComponent(option))}
-                </div>
+                {options?.map((option: AppConfigExtendedOption) => (
+                  <div
+                    key={`key_${section}_${option.name}`}
+                    className={cn(
+                      { 'w-full': option.width === 'full' },
+                      { 'w-[calc(50%-12px)]': option.width === 'half' },
+                      { 'w-[calc(33%-24px)]': option.width === 'third' },
+                      { 'w-[calc(25%-36px)]': option.width === 'quarter' },
```
michidold, line 105: use rem values, since they scale better:
```typescript
{ 'w-[calc(50%-0.75rem)]': option.width === 'half' },
{ 'w-[calc(33%-1.5rem)]': option.width === 'third' },
{ 'w-[calc(25%-2.25rem)]': option.width === 'quarter' },
 ```

michidold (changes requested):
Found some minor stuff, please have a look

michidold (approved):
Looks good

PR #345, opened 2025-01-03
307 - Allow multi selection in survey tables

maierdominik:
Issue: https://github.com/edulution-io/edulution-ui/issues/307

307:
    * save changes before testing
    * some refactoring


apps/frontend/src/components/ui/Table/ScrollableTable.tsx
```diff
@@ -40,7 +40,6 @@ interface DataTableProps<TData, TValue> {
   };
   tableIsUsedOnAppConfigPage?: boolean;
   enableRowSelection?: boolean | ((row: Row<TData>) => boolean) | undefined;
-  enableMultiRowSelection?: boolean;
```
TomlDev, line 43: Why is this removed?

apps/frontend/src/components/ui/Table/ScrollableTable.tsx
```diff
@@ -82,10 +80,10 @@ const ScrollableTable = <TData, TValue>({
     getRowId: getRowId || ((originalRow: TData) => (originalRow as { id: string }).id),
     onRowSelectionChange,
     enableRowSelection,
-    enableMultiRowSelection,
     state: {
       rowSelection: selectedRows,
     },
+    autoResetAll: false,
```
TomlDev, line 86: Why is this necessary?

maierdominik: Reverted

I have read that the change of data would reset the filter, pagination, and the sorting. I thought that is not the desired behaviour.
But I was to hasty, this could lead to further problems were we would have to add a new auto reset functionality.

apps/frontend/src/pages/Surveys/Tables/components/SurveysTablesFloatingButtons.tsx
```diff
@@ -25,59 +25,63 @@ interface SurveysTablesFloatingButtonsProps {
...
+    const ids = Object.keys(selectedRows);
+    if (ids) {
+      void deleteSurveys(ids);
       void updateUsersSurveys();
     }
   };
 
+  const isSingleSurveySelected = isExactlyOneSurveySelected();
+  const canShowResultsTable = canShowResults && (selectedSurvey?.canShowResultsTable || false);
+  const canShowResultsChart = canShowResults && (selectedSurvey?.canShowResultsChart || false);
```
TomlDev, line 56: Hä?
```
const canShowResultsTable = selectedSurvey?.canShowResultsTable && canShowResults;
const canShowResultsChart = selectedSurvey?.canShowResultsChart && canShowResults;

TomlDev (approved):

TomlDev:
The code is okay but what was missing is a confirm dialog when deleting surveys. Maybe you want to do it in another task.

maierdominik:
Created a new task were the confirmation for deletion will be established (it is on my board):
 
image

PR #347, opened 2025-01-08
Fix navigating to settings page

TomlDev:
There was a bug, for example when clicking on "manage categories" on bulletin board dropdown menu. 
grafik
You were navigated to `settings/bulletinboard` but after 0.5s you are redirected to `settings/`.

The actual bug was caused by the `<AddAppConfigDialog />` useffect, other changes are clean up.


michidold (approved):
Looks good

PR #348, opened 2025-01-08
Fix bulletin board layout issues

TomlDev:
I could not reproduce the initial problem (seems to be fixed on schulungs env). But found other issues:

Before:
grafik

So I changed the title classnames:
After:
grafik


michidold (approved):
Looks good

PR #349, opened 2025-01-08
Customize Sonner toaster

TomlDev:
### Changed

- [x] Added close button (it was not possible to change the close button's position an easy way)
- [x] Increased the offset to the border on desktop view (does not overflow user settings button anymore)
- [x] Removed unused classNames
- [x] Added a default duration to make it ALWAYS disappear after a while
- [x] Changed to a dark color (`bg-foreground`)
- [x] Enabled rich colors for the text and icon
grafik
grafik
grafik

### Unchaged

- [ ] Grouping toasts (identical toasts are not grouped to one toast, this needs an extra middleware). Maybe we can postpone it...

Since the toasts are not expanded when they are displayed I think its not so bad. Moreover this behavior should be used by us devs to track down places where this happens to refactor a multiple displaying of the toasters.

### Mobile view
I changed the colors of the content after making this screenshot. The text is white now:
grafik


michidold (approved):
Very nice

PR #350, opened 2025-01-08
Replace ImageModal with ResizableWindow component

TomlDev:
grafik


apps/frontend/src/pages/BulletinBoard/components/BulletinBoardPageColumn.tsx
```diff
@@ -57,12 +59,21 @@ const BulletinBoardPageColumn = ({
...
       </div>
 
-      {selectedImageForPreview && (
-        <ImageModal
-          isOpen={isImagePreviewModalOpen}
-          imageUrl={selectedImageForPreview}
-          onClose={closeImagePreviewModal}
-        />
+      {selectedImageForPreview && isImagePreviewModalOpen && (
+        <ResizableWindow
```
michidold, line 63: On mobile view the minimal size of the window is to large:
[Bildschirmaufzeichnung vom 2025-01-09 09-40-53.webm](https://github.com/user-attachments/assets/652ebd06-d636-4500-a4a3-c63b9fc408c5)

michidold (changes requested):
Found one small issue. Please have a look.

apps/frontend/src/pages/BulletinBoard/components/BulletinBoardPageColumn.tsx
TomlDev: As discussed I cannot reproduce it. Maybe you can find a fix for it 👍🏽

TomlDev: Should be fixed now 👍🏽

michidold (approved):
Looks good

PR #353, opened 2025-01-09
274 - Dashboard - Mail-Feed took very long with large mailbox (Testing)

maierdominik:
Issue: https://github.com/edulution-io/edulution-ui/issues/274

274:
    * mails-feed add a filter to only fetch mails from the last seven days


maierdominik:
Check if the resulting frontend and backend limits the number of mails. If not enable it!

apps/api/src/mails/mails.service.ts
```diff
@@ -104,8 +104,8 @@ class MailsService implements OnModuleInit {
       mailboxLock = await this.imapClient.getMailboxLock('INBOX');
 
       const fetchMail: AsyncGenerator<FetchMessageObject> = this.imapClient.fetch(
-        { seen: false },
-        { envelope: true, labels: true },
+        { seen: false, since: new Date(Date.now() - 7 * 24 * 60 * 60 * 1000) },
+        { limit: 10, envelope: true, labels: true },
```
TomlDev, line 108: Please change to recent true

TomlDev (approved):

PR #359, opened 2025-01-10
Fix password options for users in class management

TomlDev:
Fixed issues with restoring/setting user passwords in class management.
Additionally added a hint if the first (initial) password is currently active.


michidold (approved):
Looks good

PR #364, opened 2025-01-13
Add MobileAccessDialog to UserSettings

michidold:
If BulletinBoard is defined, the MobileAccessDialog cannot be accessed. So it is added to the UserSettings


apps/frontend/src/pages/UserSettings/MobileAccess/UserSettingsMobileAccess.tsx
```diff
@@ -0,0 +1,37 @@
...
+        iconSrc={MobileDevicesIcon}
+      />
+      <div className="mt-4 flex justify-end">
+        <Button
+          variant="btn-security"
+          size="lg"
+          onClick={() => setIsDialogOpen(!isDialogOpen)}
+        >
+          <p>{t('dashboard.mobileAccess.manual')}</p>
+        </Button>
```
TomlDev, line 25: It's currently missing the apple icon from the beginning, is that wanted?
grafik

TomlDev (approved):
👍🏽

PR #369, opened 2025-01-16
Fix redis reconnect issue

michidold:
Issue: If redis is stopped the api crashes.

Solution:
- Add reconnect feature to redis socket
- Use caching interceptor to replace caching of oidc meta data
- Use caching interceptor to cache license data

Test:
docker stop redisEdu while edulution-api is running


TomlDev (approved):
Nice 👍🏽

PR #372, opened 2025-01-20
Reduce logging for appconfig migrations

TomlDev:
Before:
grafik

---------------
After: 
grafik

------------------
With successful migration:
grafik


michidold (approved):
:+1:

PR #375, opened 2025-01-20
Sort bulletin categories

TomlDev:
I had to change the typing of the MigrationService to a T-type 👍🏽


apps/api/src/bulletin-category/bulletin-category.service.ts
```diff
@@ -170,6 +191,32 @@ class BulletinCategoryService {
     const exists = !!result;
     return { exists };
   }
+
+  async setPosition(currentUser: JWTUser, categoryId: string, newPosition: number): Promise<boolean> {
+    if (!currentUser.ldapGroups.includes(GroupRoles.SUPER_ADMIN)) {
+      throw new CustomHttpException(BulletinBoardErrorMessage.UNAUTHORIZED_UPDATE_CATEGORY, HttpStatus.FORBIDDEN);
+    }
```
michidold, line 198: You can use the `@UseGuards(AppConfigGuard)` on the Controller to check if it has permission. And that could be added to all endpoints which should be protected.

(Maybe we should change that name, since it not only protects the AppConfig)

apps/api/src/bulletin-category/bulletin-category.service.ts
```diff
@@ -170,6 +191,32 @@ class BulletinCategoryService {
...
+    if (categoryWithSamePosition) {
+      const oldPosition = category.position;
+      category.position = newPosition;
+      categoryWithSamePosition.position = oldPosition;
+      await category.save();
+      await categoryWithSamePosition.save();
+    } else {
+      category.position = newPosition;
+      await category.save();
+    }
```
michidold, line 216: You could probably use a early return 
```typescript
if (!categoryWithSamePosition) {
    category.position = newPosition;
    await category.save({ session });
    return;
  }
  ```

File: apps/frontend/src/components/ui/Table/SelectableCell.tsx [hunk 3]
```diff
@@ -0,0 +1,69 @@
...
+interface SelectableCellProps<TData> {
+  row?: Row<TData>;
+  onClick?: () => void;
+  className?: string;
+  isFirstColumn?: boolean;
+  children: React.ReactNode;
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const SelectableCell = forwardRef<HTMLDivElement, SelectableCellProps<any>>(
```
michidold, line 15: `const SelectableCell = forwardRef<HTMLDivElement, SelectableCellProps<BulletinCategoryResponseDto>>(
`
or a separate type in libs to combine every possible row type

apps/frontend/src/pages/Settings/AppConfig/bulletinboard/AppConfigBulletinCategoryTableColumn.tsx
```diff
@@ -2,13 +2,44 @@ import React from 'react';
...
+        await setCategoryPosition(id, position - 1);
+        await fetchTableContent();
+      };
+      const moveDown = async () => {
+        await setCategoryPosition(id, position + 1);
+        await fetchTableContent();
+      };
+
+      return (
+        <SortTableCell
```
michidold, line 34: This table column could be smaller

apps/frontend/src/pages/BulletinBoardEditorial/CreateOrUpdateBulletinDialogBody.tsx
```diff
@@ -58,7 +58,7 @@ const CreateOrUpdateBulletinDialogBody = ({ form }: CreateOrUpdateBulletinDialog
         <div>
           <div className="mb-1 font-bold">{t('bulletinboard.category')}</div>
           <DropdownSelect
-            options={categories}
+            options={categories?.sort((a, b) => a.position - b.position)}
```
michidold, line 61: Think it makes sense to sort it directly after fetch in the store.

michidold (changes requested):
Found some minor stuff

apps/api/src/bulletin-category/bulletin-category.service.ts
TomlDev: `return true`

apps/frontend/src/components/ui/Table/SelectableCell.tsx
TomlDev: Solved it with generics 👍🏽

apps/api/src/bulletin-category/bulletin-category.service.ts
TomlDev: https://github.com/edulution-io/edulution-ui/issues/378

PR #380, opened 2025-01-21
327 Open conference in new tab if browser disallows embedding

TomlDev:
### Changed
- Once again improved the toaster messages when starting/stopping conferences
- Added a function to test cookies
  - Checks how the browser interacts with different attributes for third party cookies and if the cookies get set
  - Opens the conference in a new tab if necessary (Chrome)
  - Opens a dialog to open the conference in a new tab because user interaction is required (Safari)
    - Chrome does not reliable return if a new tab was opened -> The dialog is being opened in Chrome too
- This PR creates a new folder `static` that will be used to host the `static/cookie-test.html` file that is used for the cookie access check.
  - The static folder will be served by Github pages once the branch is merged
  - There was no need to make an ENV variable for this since the URL must not be set by some locale admin
  - If the HTML file is not available all conferences will be opened in a new tab as fallback (Offline scenario)
  - For testing use this: https://tomldev.github.io/cookie-test/cookie-test.html

Tested with:  FF (Desktop), Chrome (Desktop), Safari (iOS), Chrome (Android)


apps/frontend/src/pages/ConferencePage/BBBFrame.tsx
```diff
@@ -1,14 +1,61 @@
...
 import useConferenceDetailsDialogStore from '@/pages/ConferencePage/ConfereneceDetailsDialog/ConferenceDetailsDialogStore';
 import { createPortal } from 'react-dom';
 import ResizableWindow from '@/components/framing/ResizableWindow/ResizableWindow';
 import { useTranslation } from 'react-i18next';
+import testCookieAccess from '@libs/common/utils/testCookieAccess';
+import LoadingIndicator from '@/components/shared/LoadingIndicator';
+import AdaptiveDialog from '@/components/ui/AdaptiveDialog';
+import { Button } from '@/components/shared/Button';
 
 const BBBIFrame = () => {
```
michidold, line 11: The file name and import in App.tsx is named BBBFrame, i would harmonize it.

libs/src/common/utils/testCookieAccess.ts
```diff
@@ -0,0 +1,51 @@
+import ExtendedCookieTestResult from '@libs/common/types/cookieTestResult';
+import COOKIE_TEST_URL from '@libs/common/constants/cookieTestUrl';
+
+async function testCookieAccess(): Promise<ExtendedCookieTestResult | null> {
```
michidold, line 4: Can you rework theses functions to arrow functions?

libs/src/common/utils/testCookieAccess.ts
```diff
@@ -0,0 +1,51 @@
...
+  try {
+    const response = await fetch(COOKIE_TEST_URL, { method: 'HEAD' });
+    if (!response.ok) {
+      return null;
+    }
+  } catch (err) {
```
michidold, line 10: `error`

michidold (approved):
Found only minor stuff.

Issue #362, opened 2025-01-10
[BUG] Survey, the group select does not seem to work


maierdominik:
When saving a created survey, the user has the possibility to add people or whole groups.

The Group select does not seem to do anything

TomlDev:
There are several bugs in the survey module. 
When I create as global-admin a survey as a anonym+public+open one, it does not appear in the list "my own surveys".
I don't find any way to update this survey (probably because it's currently only enabled on "my own surveys" page).
When I complete the survey as anonym it doesn't increase the count of attendees in the table (even after refresh).
When I complete the survey as owner it does not always update the results (table and cloud).
When I complete the survey as invited user it does not show the survey in the "already answered surveys" page.
When creating a survey it's possible to select a expiration date in the past.
When I create a survey as agy-netzint-teacher with expiration date in the past it does not show up anywhere.
When I create a survey as agy-netzint-teacher it does not show up anywhere in his pages (while it does show up for invited users).
When I click on "participate" the form is not reset from the previous fill out, despite those are different surveys.
When I then want to send my participation as invited user I get the error that I am not invited (despite the survey is visible for me).

Please, please, test the module better.

TomlDev:
Fixed in https://github.com/edulution-io/edulution-ui/pull/170

PR #170, opened 2024-09-13
Survey participation: Move to dedicated page (not inside a dialog)

maierdominik:
JIRA: https://schulz-informatik.atlassian.net/browse/NIEDUUI-394

NIEDUUI-394:
    * move survey participation into dedicated page
    * use useNavigate hook to change url parameters
        * remove surveyPageView state from stores since it is obsolete now

ADDITIONALLY:
 * Move public survey routes out of auth routes into a public routes section
     *  JIRA: https://schulz-informatik.atlassian.net/browse/NIEDUUI-392
 * Use Url params and not Search params

BEFORE:
image
image

AFTER:
image
image

PR #381, opened 2025-01-21
362 - Survey, the group select does not seem to work

maierdominik:
Issue: https://github.com/edulution-io/edulution-ui/issues/362

362:
* SaveSurveyDialog
    * fix the group select
    * fix the filters sucht that the groups of the users are respected


apps/api/src/surveys/survey-answer.service.ts
```diff
@@ -59,35 +59,34 @@ class SurveyAnswersService {
...
-            { 'invitedAttendees.username': username },
+          $or: [
+            { isPublic: true },
             {
-              $or: [
-                { $nor: [{ participatedAttendees: { $elemMatch: { username } } }] },
-                { canSubmitMultipleAnswers: true },
+              $and: [
+                {
+                  $or: [{ 'invitedAttendees.username': user.name }, { 'invitedGroups.path': { $in: user.ldapGroups } }],
```
TomlDev, line 76: This cant be correct or?
The `username` prop in JwtUser is called `prefered_username`

apps/frontend/src/locales/de/translation.json
```diff
@@ -451,7 +446,18 @@
       "participationErrorSurveyExpired": "Umfrage ist abgelaufen.",
       "idTypeError": "Ungültige Umfrage-ID",
       "missingAnswerError": "Bitte fülle die Umfrage aus bevor du deine Antwort abschickst."
-    }
+    },
+    "expires": "Läuft ab",
+    "expirationDate": "Ablaufdatum",
+    "filterPlaceHolderText": "Suche nach Umfragename",
+    "invitedAttendees": "Eingeladen",
+    "isPublic": "Öffentlich (?)",
```
TomlDev, line 454: How about 'Sichtbarkeit'

apps/frontend/src/locales/de/translation.json
```diff
@@ -451,7 +446,18 @@
...
+    "isPublicTrue": "Publik",
```
TomlDev, line 455: 'Öffentlich'

TomlDev (changes requested):
There are several bugs in the survey module.
#### anonym+public+open survey
When I create as global-admin a survey as a anonym+public+open one, it does not appear in the list "my own surveys".
I don't find any way to update this survey (probably because it's currently only enabled on "my own surveys" page).
When I complete the survey as anonym it doesn't increase the count of attendees in the table (even after refresh).
When I complete the survey as owner it does not always update the results (table and cloud).
When I complete the survey as invited user it does not show the survey in the "already answered surveys" page.

#### "normal" survey
When creating a survey it's possible to select a expiration date in the past.
When I create a survey as agy-netzint-teacher with expiration date in the past it does not show up anywhere.
When I create a survey as agy-netzint-teacher it does not show up anywhere in his pages (while it does show up for invited users).
When I click on "participate" the form is not reset from the previous fill out, despite those are different surveys.
When I then want to send my participation as invited user I get the error that I am not invited (despite the survey is visible for me).

Please, please, test the module better.

maierdominik:
For the Participation, there was another open branch, pr were some of these issues would have been fixed in (https://github.com/edulution-io/edulution-ui/pull/170)

Issue #392, opened 2025-01-23
FileSharing: Two loading animations when deleting files


TomlDev:
![Image](https://github.com/user-attachments/assets/8ef24b71-4b8f-4fff-81f2-5efcbdea382f)

Issue #387, opened 2025-01-23
OnlyOffice preview: unify the controls, use the one from the window frame


TomlDev:
![Image](https://github.com/user-attachments/assets/a8fdac78-9c62-4db0-b611-fc12950d4812)

And make it minimizable in edit mode

![Image](https://github.com/user-attachments/assets/e7cf19c0-1fc6-49dd-98f4-d0fdca60fb4d)

PR #393, opened 2025-01-23
337 Filesharing: Make the default file extension changeable

TomlDev:
### Changed
- Add documentation how to use OnlyOffice local in development
- Add OO to docker-compose file
- Set `node.env.mode === development` when using vite
- Add ODF files as templates for creating new ones
- Add switch to toggle between MSO and ODF as default file type in filesharing settings
- Refactor a lot of code
- Allow 20+ file types to be previewed with OO
- Reload AppConfigs when user navigates to settings AND reload after updating them

Use these test files:
[onlyoffice testfiles.zip](https://github.com/user-attachments/files/18522483/onlyoffice.testfiles.zip)

### Known issues
https://github.com/edulution-io/edulution-ui/issues/382
https://github.com/edulution-io/edulution-ui/issues/387
https://github.com/edulution-io/edulution-ui/issues/392


apps/frontend/src/pages/Settings/AppConfig/AppConfigPage.tsx
```diff
@@ -61,6 +61,10 @@ const AppConfigPage: React.FC = () => {
 
   const isAnAppConfigSelected = settingLocation !== '';
 
+  useEffect(() => {
+    void getAppConfigs();
+  }, []);
+
```
michidold, line 67: This is duplicate to line 56.

And can you also remove that `await getAppConfigs();` on line 156. Since we return the whole config on every change that refetch is not needed anymore.

apps/frontend/src/pages/FileSharing/fileoperations/generateFile.ts
```diff
@@ -0,0 +1,141 @@
...
+import { AvailableFileTypesType } from '@libs/filesharing/types/availableFileTypesType';
+import OnlyOfficeDocumentTypes from '@libs/filesharing/constants/OnlyOfficeDocumentTypes';
+import DocumentVendors from '@libs/filesharing/constants/documentVendors';
+import DocumentVendorsType from '@libs/filesharing/types/documentVendorsType';
+import AVAILABLE_FILE_TYPES from '@libs/filesharing/constants/availableFileTypes';
+import OPEN_DOCUMENT_TEMPLATE_PATH from '@libs/filesharing/constants/openDocumentTemplatePath';
+import { toast } from 'sonner';
+import i18n from '@/i18n';
+
+async function generateFile(
```
michidold, line 16: Can you use arrow function syntax

docker-compose.yml
```diff
@@ -30,5 +30,23 @@ services:
...
+    ports:
+      - '8088:80'
+    volumes:
+      - onlyoffice_data:/var/www/onlyoffice/Data
+      - onlyoffice_logs:/var/log/onlyoffice
+      - onlyoffice_lib:/var/lib/onlyoffice
+    environment:
+      - JWT_ENABLED=true
+      - JWT_SECRET=AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
+      - ENABLE_SSL=false
```
michidold, line 46: I would not add that here, because its not a essential container to run edulution. I would use the new docker integration to create a container. Actually, I am finalizing it, so it should be a one click install.

README.md
michidold, line 1: I would not add that here, because its not a essential container to run edulution. I would use the new docker integration to create a container. Actually, I am finalizing it, so it should be a one click install. The documentation will then be on doku.edulution.io

michidold (changes requested):
Found only some minor stuff.

docker-compose.yml
TomlDev: Ah I thought this docker compose is only for development 👍🏽  Removed

README.md
TomlDev: Is the readme only for production? Because for development I think this is still necessary.
Or you mean to only to adjust this to use the new docker integration? Have done it now

michidold (approved):
Looks good

PR #395, opened 2025-01-23
Fix proxy path generation and add form validation

michidold:
- Added form resolver schema to appconfig page (not finalized yet)
- Prohibit the usage of ui routes as proxy path in traefik config
- Rearranged mails config (ToDo: Rework as extended appconfig table) 
  -  Task: https://github.com/edulution-io/edulution-ui/issues/400
- Smaller bugfixes like default values to form fields


TomlDev (approved):
I fixed what I found 👍🏽

PR #403, opened 2025-01-28
Replace token in native framed app urls

TomlDev:
Also fixes that appconfigs are not reloaded when one is updated


michidold (approved):
Looks good

PR #305, opened 2024-12-10
286 Add veyon webapi to classmgmt

michidold:
Setup Veyon: https://doku.netzint.de/doku.php?id=intern:technik:schulserver:edulution:guides:veyon

- Added veyon module to edu-api
- Added framebufffer image to UserCard component
- Reworked appconfig module in edu-api
- Reworked AppConfigTable in a more generic way

To do's in a separate task:

- https://github.com/edulution-io/edulution-ui/issues/357
- https://github.com/edulution-io/edulution-ui/issues/358
- https://github.com/edulution-io/edulution-ui/issues/352

PR #406, opened 2025-01-29
286 Add veyon api support to class management

michidold:
Fixes review findings from:
https://github.com/edulution-io/edulution-ui/pull/305


apps/api/src/appconfig/appconfig.controller.ts
TomlDev, line 1: Is there a reason why you did not change from `name` identifier for appConfigs to `id`?

apps/frontend/src/pages/Settings/AppConfig/classmgmt/AddVeyonProxyDialog.tsx
TomlDev, line 1: Folder is still called `classmgmt`

apps/api/src/appconfig/appconfig.controller.ts
TomlDev, line 1: Did you create some documentation on how to add a table to the settings?
https://github.com/edulution-io/edulution-ui/pull/305#discussion_r1918742023

apps/api/src/appconfig/appconfig.service.ts
```diff
@@ -91,6 +91,13 @@ class AppConfigService implements OnModuleInit {
     }
   }
 
+  static checkIfFileExistAndDelete(filePath: string) {
+    if (existsSync(filePath)) {
+      unlinkSync(filePath);
+      Logger.log(`${filePath} deleted.`, AppConfigService.name);
+    }
+  }
```
TomlDev, line 99: This belongs in the filesystem service

apps/api/src/veyon/veyon.service.ts
```diff
@@ -35,23 +36,38 @@ class VeyonService implements OnModuleInit {
...
+      if (!appConfig.extendedOptions || Object.keys(appConfig.extendedOptions).length === 0) {
         return;
       }
 
       // ToDo: Add support for more proxies https://github.com/edulution-io/edulution-ui/issues/352
       const veyonProxies = appConfig.extendedOptions.VEYON_PROXYS as VeyonProxyItem[];
-      const veyonApiUrl = veyonProxies[0].proxyAdress;
+
+      if (!Array.isArray(veyonProxies) || veyonProxies.length === 0) {
+        return;
```
TomlDev, line 48: Does it make sense to just return here and in line 41 without returning an error?

apps/frontend/src/pages/Settings/AppConfig/appConfigSchema.ts
TomlDev, line 1: For me it was not possible to save the AppConfig:
grafik
When I pressed `save` nothing happened and on the right side you can see that there is some required field in the form that causes the error.

apps/frontend/src/pages/Settings/AppConfig/appConfigSchema.ts
```diff
@@ -1,42 +1,48 @@
...
 const forbiddenRouts = [...Object.values(APPS), 'auth', 'edu-api'];
 
-const appConfigFormSchema = z.record(
-  z.object({
-    appType: appIntegrationEnum,
-    accessGroups: z.array(z.object({})).optional(),
-    options: z.object({
-      url: z.string().optional(),
-      apiKey: z.string().optional(),
+const appConfigFormSchema = (t: TFunction<'translation', undefined>) =>
```
TomlDev, line 11: Since this is a function now I would prefer to call it `getAppConfigFormSchema`

apps/frontend/src/pages/ClassManagement/LessonPage/UserArea/FrameBufferImage.tsx
```diff
@@ -15,24 +17,20 @@ type FrameBufferImageProps = {
 const FrameBufferImage: React.FC<FrameBufferImageProps> = ({ user }) => {
   const { t } = useTranslation();
   const [imageSrc, setImageSrc] = useState<string>('');
-  const { authenticateVeyonClients, getFrameBufferStream } = useVeyonApiStore();
-  const [connUid, setConnUid] = useState<string>('');
+  const { userConnectionUids, authenticateVeyonClient, getFrameBufferStream } = useVeyonApiStore();
   const [isImagePreviewModalOpen, setIsImagePreviewModalOpen] = useState(false);
 
+  const connectionUid = userConnectionUids.find((conn) => conn.veyonUsername === user.name)?.connectionUid || '';
```
TomlDev, line 23: Is this comparison correct? It uses the non-unique value `user.name`

apps/frontend/src/pages/ClassManagement/LessonPage/UserArea/UserArea.tsx
```diff
@@ -97,7 +97,10 @@ const UserArea = () => {
           />
         ))}
       </div>
-      {isMemberSelected ? <LessonFloatingButtonsBar students={getSelectedStudents()} /> : null}
+      <LessonFloatingButtonsBar
+        students={getSelectedStudents()}
+        isMemberSelected={!!isMemberSelected}
```
TomlDev, line 102: Could you please do me a favor and rename this `isMemberSelected` in the component to `selectedMemberCount`

apps/frontend/src/utils/handleApiError.ts
```diff
@@ -11,13 +11,23 @@ import CustomAxiosError from '@libs/error/CustomAxiosError';
...
     const errorMessage = i18n.t(axiosError.response?.data?.message) || axiosError.response?.statusText;
 
-    toast.error(errorMessage || i18n.t('errors.unexpectedError'));
+    if (!displayedErrors.has(errorMessage)) {
+      displayedErrors.add(errorMessage);
+      toast.error(errorMessage || i18n.t('errors.unexpectedError'));
+
+      setTimeout(() => {
+        displayedErrors.delete(errorMessage);
+      }, 5000);
```
TomlDev, line 28: 👍🏽 

I think we should use the same value as the toaster duration and therefore maybe extract it to a const and use on both places `apps/frontend/src/components/ui/Sonner.tsx:19`

TomlDev (changes requested):
I could not fully test it since it did not allow me to add the AppConfig anymore (see comment)

apps/api/src/appconfig/appconfig.controller.ts
michidold: Wanted to resolve that in another issue. Maybe we can address that in:
https://github.com/edulution-io/edulution-ui/issues/430

apps/frontend/src/pages/Settings/AppConfig/classmgmt/AddVeyonProxyDialog.tsx
michidold: Should be a shortcut, but since we have these specifig domains i think you're right. I'll rename it.

apps/api/src/appconfig/appconfig.controller.ts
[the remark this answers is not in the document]
TomlDev: Do you know why it's not accessible for me?

TomlDev (approved):
👌🏽

PR #412, opened 2025-01-30
411 Onlyoffice bug when opening preview

TomlDev:
I had to add
```
allowedHosts: ['host.docker.internal'],
```
to the vite config, otherwise it didn't allow to call that URL.


michidold (approved):
Works, delete files after opening will be fixed in a separate branch.

PR #418, opened 2025-01-31
Fix mail sync fetch via email

michidold:
- Load pubkey to memory
- Add get email address decorator
- Use email for all sync job queries
- Add tooltip with execution state

### DIFF
--- a/apps/api/src/common/decorators/getUser.decorator.ts
+++ b/apps/api/src/common/decorators/getUser.decorator.ts
@@ -19,4 +19,12 @@ export const GetCurrentUsername = createParamDecorator((_data: unknown, ctx: Exe
   return request.user.preferred_username;
 });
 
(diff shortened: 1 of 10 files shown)


apps/api/src/common/decorators/getUser.decorator.ts
```diff
@@ -19,4 +19,12 @@ export const GetCurrentUsername = createParamDecorator((_data: unknown, ctx: Exe
   return request.user.preferred_username;
 });
 
+export const GetUsersEmailAddress = createParamDecorator((_data: unknown, ctx: ExecutionContext): string => {
+  const request: Request = ctx.switchToHttp().getRequest();
+  if (!request.user?.email) {
+    throw new UnauthorizedException('Email in JWT is missing');
+  }
+  return request.user.email;
+});
```
TomlDev, line 28: please move that to its own file, that this file has multiple exports was a mistake

TomlDev (approved):
👌🏽

PR #420, opened 2025-01-31
Fix upload files dialog

TomlDev:


apps/frontend/src/pages/FileSharing/utilities/UploadContentBody.tsx
```diff
@@ -61,7 +59,8 @@ const DropZone: FC<DropZoneProps> = ({ files, setFiles }) => {
             <MdOutlineCloudUpload className="h-12 w-12 text-muted" />
           </div>
         ) : (
-          <p className="font-bold text-ciRed">
+          <p className="inline-flex items-center font-bold text-ciRed">
+            <HiExclamationTriangle className="mr-2 h-6 w-6" />
             {fileUploadSize > MAX_FILE_UPLOAD_SIZE
```
michidold, line 64: The upload limitation is per file, so technically it should be able to upload more files, if they are smaller than 50MB each

apps/frontend/src/pages/FileSharing/utilities/UploadContentBody.tsx
```diff
@@ -78,52 +77,47 @@ const DropZone: FC<DropZoneProps> = ({ files, setFiles }) => {
...
+                </div>
+              )}
+              <Button
+                onClick={() => removeFile(file.name)}
+                className="absolute right-1 top-1 h-8 rounded-full bg-red-500 bg-opacity-70 p-2"
+              >
+                <HiTrash className="text-text-ciRed h-4 w-4 hover:text-red-700" />
+              </Button>
+
+              <TooltipProvider>
```
michidold, line 109: The tooltip is cut off:
grafik

apps/frontend/src/pages/FileSharing/utilities/UploadContentBody.tsx
TomlDev: I have created a task for this:
https://github.com/edulution-io/edulution-ui/issues/424

apps/frontend/src/pages/FileSharing/utilities/UploadContentBody.tsx
TomlDev: Good catch. I removed the tooltip now and replaced with this behavior:

https://github.com/user-attachments/assets/e3ab4d50-f3f9-4f27-978a-abada64d2e66
It's not the cleanest solution but for the moment it should work.

michidold (approved):
Looks good

PR #427, opened 2025-02-03
Add licence file and headers

michidold:
- add script to automatically add headers on commit
- add license file and exception
- updated all ts/tsx/js/jsx files exept for *.config.ts/js


apps/api-e2e/src/api/api.spec.ts
TomlDev, line 1: Would it make sense to also mention the license exception file?

TomlDev (approved):
Looks good 👍🏽

PR #500, opened 2025-02-21
484 fix npm audit issue

michidold:


apps/api/.env.production
TomlDev, line 1: Was this changed on purpose? Is it unused?

.eslintrc.json
```diff
@@ -13,25 +13,16 @@
...
-    "react-refresh",
-    "@typescript-eslint",
-    "import",
-    "prettier",
-    "jsx-a11y",
-    "react-hooks",
-    "n",
-    "promise"
-  ],
+  "plugins": ["react", "@typescript-eslint", "import", "prettier", "jsx-a11y"],
```
TomlDev, line 25: "n" and "promise" are also removed, is this on purpose?

TomlDev (changes requested):
Looks good 👍🏽

.eslintrc.json
TomlDev: It is unused since no rules are imported

TomlDev (approved):

PR #504, opened 2025-02-24
Github: Move cookie test file to folder that can be used for GH pages build

TomlDev:
Should fix conferences opening **always** in new tab.


michidold (approved):
Hard to test, but think it may work

Issue #523, opened 2025-02-25
FileSharing: Missing translation for long filenames


TomlDev:
![Image](https://github.com/user-attachments/assets/2fc286ed-43a8-4c47-91b9-fd448a849b31)

PR #532, opened 2025-02-25
Filesharing: Add German / English translation for maximum character limit

Maddi02:


TomlDev (approved):
Probably closes https://github.com/edulution-io/edulution-ui/issues/523

PR #549, opened 2025-02-28
Replace loading indicator in filesharing

TomlDev:
Changed from the loading spinner to a horizontal line because the old one made the layout flip and led to problems with the new preview.

48 Files were changed because of file moves and a renaming of the LoadingIndicator to LoadingIndicatorDialog.


apps/frontend/src/components/ui/Table/ScrollableTable.tsx
```diff
@@ -110,7 +110,7 @@ const ScrollableTable = <TData, TValue>({
 
   return (
     <>
-      {isLoading && data?.length === 0 ? <LoadingIndicator isOpen={isLoading} /> : null}
+      {isLoading && data?.length === 0 ? <LoadingIndicatorDialog isOpen={isLoading} /> : null}
```
michidold, line 113: Would prefer using here the CircleLoader without a black background

michidold (changes requested):
One small question

apps/frontend/src/components/ui/Table/ScrollableTable.tsx
TomlDev: I don't think that it makes sense here. The black background is because the Loader is overlaying the content.
It also visualizes that the content is blocked in this moment. If we remove the background we should not simply show it in the same position (because it looks weird when it's directly displayed on text).

And since your change is not related to this PR I would say we should have an extra task for this...

michidold: Yes, but in a lot of cases its shown instead of the content. We should definitely address this in a separate task. To unify this. 

I've added one: https://github.com/edulution-io/edulution-ui/issues/571

michidold (approved):

PR #550, opened 2025-02-28
Fix opening the same file multiple times

TomlDev:
Before this PR when you switched the file for preview fast you had many errors and files won't load.
This PR introduces an AbortController that cancels the requests.


apps/frontend/src/pages/FileSharing/FilePreview/FileRenderer.tsx
```diff
@@ -34,8 +35,26 @@ const FileRenderer: FC<FileRendererProps> = ({ editMode }) => {
...
+    }
+  }, [isEditorLoading, isDownloadFileLoading, isGetDownloadLinkUrlLoading, currentlyEditingFile?.basename]);
+
+  useEffect(
+    () => () => {
+      if (currentlyEditingFile) {
+        void setFileIsCurrentlyDisabled(currentlyEditingFile.basename, false);
+      }
+    },
+    [currentlyEditingFile?.basename],
```
michidold, line 56: This is some kind of double code, maybe could be unified:

```typescript
useEffect(() => {
  if (currentlyEditingFile && !isEditorLoading && !isDownloadFileLoading && !isGetDownloadLinkUrlLoading) {
    void setFileIsCurrentlyDisabled(currentlyEditingFile.basename, false);
  }

  return () => {
    if (currentlyEditingFile) {
      void setFileIsCurrentlyDisabled(currentlyEditingFile.basename, false);
    }
  };
}, [currentlyEditingFile, isEditorLoading, isDownloadFileLoading, isGetDownloadLinkUrlLoading]);
```

apps/api/src/filesharing/filesharing.controller.ts
```diff
@@ -114,14 +114,22 @@ class FilesharingController {
...
+      if (err.message !== 'Premature close') {
+        Logger.error(`${FilesharingController.name}: Unknown stream error:`, err);
+      }
     });
+
+    const fileName = filePath.split('/').pop();
+
+    res.setHeader('Content-Disposition', `attachment; filename="${fileName}"`);
+
+    stream.pipe(res);
```
michidold, line 132: Normally we handle logic only in the service it self, could it be moved to filesharing service as well?

michidold (changes requested):
Some code improvements only

apps/api/src/filesharing/filesharing.controller.ts
TomlDev: The AI says this, do you still want me to change it?

**Short Answer**  
It can make sense to move more of the streaming logic to the service if that logic is truly “business logic.” However, in many NestJS applications, it’s also common (and perfectly acceptable) for the controller to set response headers and pipe the returned stream to the HTTP response. Whether you move it all into the service depends on how strictly you want to keep the controller “thin” versus how practical it is for the service to manage the streaming details.

---

## Why the review comment?

The comment “Normally we handle logic only in the service” points to a common NestJS recommendation:  
- **Controllers** deal with request/response handling (routing, extracting parameters, setting status codes, etc.).  
- **Services** contain business or application logic (e.g., retrieving data from a database, processing files, etc.).

This pattern keeps controllers as lightweight as possible.

---

## Where the streaming logic fits

1. **Service**  
   - Retrieving the file or stream from a storage location or WebDAV server.  
   - Handling any application/business logic for file permissions, access, or transformations.

2. **Controller**  
   - Setting **HTTP headers** like `"Content-Disposition"`.  
   - Handling **request parameters** (e.g., `filename` from the query).  
   - **Piping the stream** to the user’s response.

Conceptually, “_stream the file to the user_” can be seen as a part of delivering an HTTP response (thus, controller responsibility). On the other hand, the logic around _how_ you retrieve that stream from a remote location (the WebDAV server) can be service-level logic.

---

## Possible Approaches

### 1. Keep streaming details in the controller  
Your current approach has the service return a stream, and the controller does:

```ts
@Header('Content-Type', 'application/octet-stream')
async webDavFileStream(
  @Query('filePath') filePath: string,
  @Res() res: Response,
  @GetCurrentUsername() username: string,
) {
  const stream = await this.filesharingService.getWebDavFileStream(username, filePath);
  stream.on('error', (err) => { ... });
  res.setHeader('Content-Disposition', `attachment; filename="${fileName}"`);
  stream.pipe(res);
}
```

**Pros**  
- Clear that the controller handles outgoing response details.  
- The service only needs to focus on how to obtain the stream.

**Cons**  
- The controller might grow if you have more advanced streaming logic (retry, chunking, partial content, etc.).

### 2. Move more logic into the service  
You could do something like:

```ts
async webDavFileStream(@Query('filePath') filePath: string, ...): Promise<Stream> {
  return this.filesharingService.createDownloadStream(username, filePath);
}
```

Where `createDownloadStream` might even handle errors, set recommended HTTP headers, and so on. The controller simply returns the result. However, this typically couples your service to NestJS’s `Response` object if you want the service itself to set headers – not always ideal if you want your service to remain framework‑agnostic.

---

## Which approach to choose?

- If you prefer a strictly “thin” controller, you could push almost everything into the service and have the controller do minimal work. But you’ll need to decide how the service communicates HTTP-specific metadata (like headers) back to the controller.
- If you’re okay with the controller handling “response shaping” (headers, stream piping), then you can keep it as it is. This is a _very common_ and reasonable pattern in NestJS.

**In other words, it’s not strictly wrong** to keep the streaming in your controller if that’s the layer where you’re managing HTTP response details. The comment is likely suggesting a more purist separation of concerns, but many production NestJS apps handle streaming exactly in the controller. 

---

## Conclusion

- **Yes, it can make sense** to move more of the streaming logic to the service if you want your controller to remain ultra-thin.  
- **No, it’s not necessarily wrong** to handle the streaming in the controller, because setting headers and piping the stream is often considered part of HTTP response handling—which is exactly what the controller is for.  

Ultimately, it’s a design choice. If your team or codebase follows a strict pattern of *all logic in the service*, then you might refactor. Otherwise, your current approach is also valid.

apps/frontend/src/pages/FileSharing/FilePreview/FileRenderer.tsx
TomlDev: No, this cannot be done without changing the logic, the AI says this:

> Generally, **no**—you can’t *cleanly* merge those two `useEffect` hooks without changing their behavior. Each hook handles a distinct scenario:
> 
> 1. **The first** effect runs whenever any of these change:
>    - `isEditorLoading`
>    - `isDownloadFileLoading`
>    - `isGetDownloadLinkUrlLoading`
>    - `currentlyEditingFile?.basename`
> 
>    and if certain conditions are met (no loading), it re‑enables the file.
> 
> 2. **The second** effect is a cleanup effect. It only depends on `currentlyEditingFile?.basename` and re‑enables the file on unmount *or* when the file reference changes.
> 
> If you tried to unify them into one effect with a combined dependency array, the cleanup function would also trigger on every change in `isEditorLoading`, `isDownloadFileLoading`, or `isGetDownloadLinkUrlLoading`, which is different from the current behavior. That could cause re‑enabling the file at unintended times, thus altering the logic.
> 
> In other words, the second effect is specifically for “tear down” (unmount or new file), while the first is for “ready state.” Merging them into a single hook would change the conditions under which the cleanup runs. So, it’s simpler—and safer—to keep them separate.

michidold (approved):

PR #552, opened 2025-03-01
Only Office: Created tables and presentations in OpenDocument format are not readable

Maddi02:


Maddi02:
As you can see in the picture, the file size is now correct. It was just the wrong path being declared. 

image

Maddi02:
[This](https://github.com/edulution-io/edulution-ui/pull/552#pullrequestreview-2653592166) should be done

TomlDev (approved):

PR #554, opened 2025-03-03
Add initial sorting support to ScrollableTable and FileSharingTable

Maddi02:
Introduced an `initialSorting` prop in ScrollableTable to allow default sorting configuration. Updated FileSharingTable to utilize this feature for predefined sorting by 'type' and 'select-filename'. This improves table usability by setting a consistent initial sort order.

The sorting problem mentioned in the issue should be fixed. 

image


apps/frontend/src/components/ui/Table/ScrollableTable.tsx
```diff
@@ -68,13 +69,19 @@ const ScrollableTable = <TData, TValue>({
...
   const { t } = useTranslation();
 
   const hasPositionCol = useMemo(() => columns.some((c) => c.id === 'position'), [columns]);
 
-  const [sorting, setSorting] = useState(hasPositionCol ? [{ id: DEFAULT_TABLE_SORT_PROPERTY_KEY, desc: false }] : []);
+  const [sorting, setSorting] = useState(() => {
+    if (initialSorting && initialSorting.length > 0) {
+      return initialSorting;
+    }
+    return hasPositionCol ? [{ id: DEFAULT_TABLE_SORT_PROPERTY_KEY, desc: false }] : [];
```
TomlDev, line 83: Could `hasPositionCol` be removed with your changes and replaced with the `initialSorting` logic where it was used?

TomlDev (changes requested):
Nice and clean solution 👍🏽

apps/frontend/src/components/ui/Table/ScrollableTable.tsx
Maddi02: Yes, we can remove hasPositionCol if we always pass initialSorting explicitly. However, because that would require forcing consumers to provide this prop for any fallback sorting, I’d keep it as is for now.

TomlDev: As discussed it's not that easy because the `position` field is used in the `AppConfigBulletinCategoryTable` which uses the generalized table component.

TomlDev (approved):

PR #568, opened 2025-03-04
Translate backend error when request is too large

TomlDev:
### Changed
- Some linting
- Refactor error toaster when survey is too big
- Add default translation when backend responds with HTTP status 413
-


michidold (approved):
Looks good

PR #570, opened 2025-03-04
Fix zIndex of sidebar

TomlDev:
This was a general problem, not only related to the newest FileSharing changes.

grafik


michidold (approved):
Looks good

PR #585, opened 2025-03-06
489 - Settings navigation fails after creating redirect app

michidold:
- Add useEffect to AddAppConfigDialog to navigate when submit is done
- Cleanup missing name in deleteAppConfig toast
- Add useEffect to MenuBar to update isSelected menu bar entry
- Renamed survey menu bar entry ids to work correctly


TomlDev (approved):
Good job. 👍🏽 
I think you already wanted to fix the menubar also for surveys, but that part is not working. I would suggest to do it in this PR because it's closely related.
I would go with an approach where we check if the `part[1]` of the `path` exists in the `menuItems` and if so we set it.

TomlDev (approved):
Nice 👍🏽

PR #588, opened 2025-03-06
587 - Back button does not work in classmgmt and survey pages

michidold:
Add `replace ` to the <Navigate /> operation to prevent a new history entry and enable the back and forth click


TomlDev (approved):
👍🏽

Issue #578, opened 2025-03-05
Survey Creator: Remove sync between surveyjs and form


TomlDev:
Currently we update our form every time the user changes something in the survey.
The state of the survey should only be used once: when the survey is saved.
And when the survey is initialized to edit it.

PR #589, opened 2025-03-06
578 - Survey Creator: Remove sync between surveyjs and form

maierdominik:
Issue: https://github.com/edulution-io/edulution-ui/issues/578

578:
* split the former 'SurveyEditor' in two 
    * one function for the rendering (the new 'SurveyEditor')
    * and one for the `updateOrCreateSurveyCreatorModel` component that is rendered
* move the `SaveSurveyDialog` and the `FloatingButtonsBar` into the new render component
* instead of passing the `form` as prop we now pass each `getter` and `setter` to the child compoentents


maierdominik:
duplicated

We decided to go with another approach were we use a `ref` in order to prevent this many props getting passed to each other

PR #593, opened 2025-03-07
578 - Remove sync between surveyjs and form

TomlDev:
Issue: https://github.com/edulution-io/edulution-ui/issues/578


michidold (changes requested):
At the moment the survey is lost, if the site is refreshed or location changes. survey-js offers a manual to store it in the localStorage:
https://surveyjs.io/form-library/documentation/how-to-save-and-restore-incomplete-survey
We should then clear this when we clearAllStores

michidold (approved):
:+1:

Issue #558, opened 2025-03-03
Survey Creator: Analyse why the expiration date does reset with backspace (sometimes)


maierdominik:
Sometimes does the backspace remove the complete expiration date when only the
value of the hours should get removed.

![Image](https://github.com/user-attachments/assets/ef1022ad-5b4b-49a9-b8e1-6962dc3dc631)

Not just caused by the wrong time layout that can be fixed with:
```
convertDateToDateTimeInput(date: Date): string {
  const year = String(date.getFullYear()).padStart(4, '0');
```

TomlDev:
![Image](https://github.com/user-attachments/assets/20826202-6121-4ade-83a0-083f605fa5bc)

It crashes with this survey data:
```
{"_id":"67adbb5c4169f5ce0f5e6fa1","formula":{"title":"Neue Umfrage","pages":[{"name":"Seite1","elements":[{"type":"text","name":"Frage1"}]}]},"backendLimiters":[],"saveNo":1,"creator":{"firstName":"Global","lastName":"Admin","username":"global-admin","_id":"67c59990c4e74a5b96ae4325","id":"67c59990c4e74a5b96ae4325"},"invitedAttendees":[],"invitedGroups":[],"participatedAttendees":[],"answers":["67af11deaf8647352d29ee00","67af133aebe2d7d91ebe2c95","67af13968ebd4f8bd363ef07"],"isAnonymous":true,"isPublic":true,"canSubmitMultipleAnswers":true,"createdAt":"2025-02-13T09:28:37.170Z","updatedAt":"2025-03-03T11:59:12.593Z","__v":0,"canUpdateFormerAnswer":true,"schemaVersion":1,"expires":"2025-07-14T09:11:00.000Z","id":"67adbb5c4169f5ce0f5e6fa1"}

PR #596, opened 2025-03-07
558 - Survey Creator: Analyse why the expiration date does reset with backspace (sometimes)

maierdominik:
Issue: https://github.com/edulution-io/edulution-ui/issues/558

558:
    * add the DateTimePicker component from the shadcn/ui-expansions (https://shadcnui-expansions.typeart.cc/docs/datetime-picker) (OPEN SOURCE)
        * therefore we add the Select component from shadcn
        * fix all the eslint errors
        * use the already existing calendar component
        * adapt the color/variant (using the DropdownVariants and Styles)
        * use the minDate to disable previous dates


apps/frontend/src/components/shared/DateTimePicker/convertDateToDateTimeInput.ts
```diff
@@ -10,16 +10,27 @@
...
+  const year = addLeadingZero(date.getUTCFullYear(), 4);
   const month = addLeadingZero(date.getMonth() + 1);
   const day = addLeadingZero(date.getDate());
   const hours = addLeadingZero(date.getHours());
   const minutes = addLeadingZero(date.getMinutes());
 
   return `${year}-${month}-${day}T${hours}:${minutes}`;
 }
+
+export default function fixDateTimeInput(value: string | Date | undefined): string {
```
TomlDev, line 27: Please rename something here. The function is imported as `convertDateToDateTimeInput` and the file has the same name.

TomlDev (changes requested):
The issue is still there:

https://github.com/user-attachments/assets/dcd9506a-f795-4637-884b-4e06e54ea2df

apps/frontend/src/components/ui/DropdownSelect/DropdownSelect.tsx
```diff
@@ -21,7 +21,7 @@ export type DropdownOptions = {
   name: string;
 };
 
-type DropdownVariant = 'dialog' | 'default';
+export type DropdownVariant = 'dialog' | 'default';
```
TomlDev, line 24: Please move it to the shared lib then

apps/frontend/src/components/ui/DateTimePicker.tsx
TomlDev, line 1: Is this the component that was partially installed from shadcn extensions? If we should only use the wrapper `apps/frontend/src/components/ui/DatetimePickerHourCycle.tsx` component I would rename this to `SH` and move it to the folder.

libs/src/survey/constants/initial-survey-form.ts
```diff
@@ -24,7 +24,7 @@ const getInitialSurveyFormValues = (creator: AttendeeDto, selectedSurvey?: Surve
   participatedAttendees: selectedSurvey?.participatedAttendees || [],
   answers: selectedSurvey?.answers || [],
   createdAt: selectedSurvey?.createdAt || new Date(),
-  expires: selectedSurvey?.expires,
+  expires: selectedSurvey?.expires ? new Date(selectedSurvey?.expires) || undefined : undefined,
```
TomlDev, line 27: > Functionally, it works, but the `|| undefined` part is redundant because `new Date(...)` will always return a `Date` object—even if the input is invalid (in which case it’s an “Invalid Date” object, but still truthy). So effectively:
> 
> ```ts
> new Date(selectedSurvey?.expires) || undefined
> ```
> 
> will just be `new Date(selectedSurvey?.expires)` every time. You could safely simplify this to:
> 
> ```ts
> expires: selectedSurvey?.expires
>   ? new Date(selectedSurvey?.expires)
>   : undefined,
> ```
> 
> If your intention was to yield `undefined` when the date is invalid, you’d have to explicitly check for that scenario, e.g.:
> 
> ```ts
> const expiresDate = new Date(selectedSurvey?.expires);
> expires: !isNaN(expiresDate.getTime()) ? expiresDate : undefined;
> ```
> 
> Otherwise, an invalid date will produce an “Invalid Date” object rather than `undefined`.

apps/frontend/src/pages/BulletinBoard/BulletinBoardEditorial/CreateOrUpdateBulletinDialogBody.tsx
```diff
@@ -109,14 +109,14 @@ const CreateOrUpdateBulletinDialogBody = ({ form }: CreateOrUpdateBulletinDialog
 
           {isActive && !isPermanentlyActive && (
             <>
-              <DateAndTimeInput
-                form={form}
-                name="isVisibleStartDate"
+              <DatetimePickerHourCycle
+                value={form.watch('isVisibleStartDate')}
+                onChange={(value) => form.setValue('isVisibleStartDate', value)}
```
TomlDev, line 114: In `apps/frontend/src/pages/Surveys/Editor/dialog/SaveSurveyDialogBody.tsx` you pass a variant, both are dialogs. Where is the difference?

apps/frontend/src/components/ui/DatetimePickerHourCycle.tsx
TomlDev, line 1: If you renamed `apps/frontend/src/components/ui/DateTimePicker.tsx` to an `SH` component I would propose to rename this to simply `DateTimePicker` to make it clear that this component should be used.

TomlDev (changes requested):
Some small findings 👍🏽

libs/src/survey/constants/initial-survey-form.ts
maierdominik: Since i get an error when passing `undefined` to `new Date()` I had to add another check for a defined value.

ERROR:
image

RESOLVED:
image

apps/frontend/src/components/ui/DatetimePickerHourCycle.tsx
maierdominik: To make it more obvious we moved the new `DateTimePicker.tsx` (former `DatetimePickerHourCycle.tsx`) to the Shared section `apps/frontend/src/components/shared`

apps/frontend/src/pages/BulletinBoard/BulletinBoardEditorial/CreateOrUpdateBulletinDialogBody.tsx
maierdominik: Thx, I forgot to add it.
After, i styled the component for the Survey Creator.

TomlDev (changes requested):
For me with DE as language it was still in EN
Also I had a delay when inserting date/time of 1 second which is not acceptable
And I don't see the advantage over the native datetime-local input field at the moment (except styling)

maierdominik:
This PR is deprecated there is a new one on: https://github.com/edulution-io/edulution-ui/pull/613

PR #597, opened 2025-03-07
401 - Settings - Reset yaml if proxy path is cleared

michidold:
- Add extra descriptions
- Rework component to load either saved config, or a template.
- edit fields now enabled with expert mode

Known issue -> setValue inside a child component leads to a console.error in the parent component if the useForm state is handled in both components. Found no useful solution so far. Inputs welcome


apps/frontend/src/pages/Settings/AppConfig/components/ProxyConfigForm.tsx
```diff
@@ -34,179 +38,224 @@ type ProxyConfigFormProps = {
...
+                    <YamlEditor
+                      value={field.value}
+                      onChange={field.onChange}
+                      disabled={!expertModeEnabled}
+                    />
+                  </FormControl>
+                  <FormMessage className="text-p" />
+                </FormItem>
+              )}
+            />
```
TomlDev, line 236: You should use our `<FormField` component instead

apps/frontend/src/pages/Settings/AppConfig/components/ProxyConfigForm.tsx
```diff
@@ -34,179 +38,224 @@ type ProxyConfigFormProps = {
...
+            <div className="flex flex-row justify-end">
+              <Button
+                className="mr-4"
+                type="button"
+                variant="btn-infrastructure"
+                size="lg"
+                onClickCapture={handleLoadDefaultConfig}
```
TomlDev, line 243: Is there a special reason why you use `onClickCapture` instead of `onClick?`?

TomlDev (changes requested):
Looks good, please have a look 👍🏽

apps/frontend/src/pages/Settings/AppConfig/components/ProxyConfigForm.tsx
michidold: Actually not, in the former implementation something hasn't worked with onClick itself. Now, it seems to work. Renamed it.

apps/frontend/src/pages/Settings/AppConfig/components/ProxyConfigForm.tsx
michidold: Replaced the first two input fields. Switch and YamlEditor are kept

TomlDev (approved):
👍🏽

PR #602, opened 2025-03-10
[Filesharing] Fix API file type path URL formatting

Maddi02:
by adding a leading slash to the path parameter


TomlDev (approved):
I don't know why but it works.

Issue #560, opened 2025-03-03
Survey: The creator should only reset when it was saved and the user was navigated


TomlDev:
At the moment the survey is lost, if the site is refreshed or location changes. survey-js offers a manual to store it in the localStorage:
https://surveyjs.io/form-library/documentation/how-to-save-and-restore-incomplete-survey
We should then clear this when we clearAllStores

PR #607, opened 2025-03-11
560 - Survey: The creator should only reset when it was saved and the user was navigated 

maierdominik:
Issue: https://github.com/edulution-io/edulution-ui/issues/560

560:
* make the store persistent such that the changes of a survey are stored when moving in edulution
* prevent the page from refreshing because the values would get lost


libs/src/survey/constants/initial-survey-form.ts
```diff
@@ -14,21 +14,32 @@ import i18next from 'i18next';
...
+    participatedAttendees: storedSurvey?.participatedAttendees || selectedSurvey?.participatedAttendees || [],
+    answers: selectedSurvey?.answers || selectedSurvey?.answers || [],
+    createdAt: selectedSurvey?.createdAt || selectedSurvey?.createdAt || new Date(),
+    expires: expiresDate && !Number.isNaN(expiresDate.getTime()) ? expiresDate : undefined,
+    isAnonymous: selectedSurvey?.isAnonymous || false,
+    canSubmitMultipleAnswers: selectedSurvey?.canSubmitMultipleAnswers || false,
+    isPublic: selectedSurvey?.isPublic || false,
+    canUpdateFormerAnswer: selectedSurvey?.canUpdateFormerAnswer || false,
+  };
+};
```
TomlDev, line 43: Unrelated changes?

apps/frontend/src/pages/Surveys/Editor/SurveyEditorPage.tsx
```diff
@@ -35,17 +35,29 @@ import LoadingIndicatorDialog from '@/components/ui/Loading/LoadingIndicatorDial
...
+    resetStoredSurvey,
+  } = useSurveyEditorPageStore();
+
   const { t } = useTranslation();
   const navigate = useNavigate();
   const { user } = useUserStore();
   const { surveyId } = useParams();
   const { language } = useLanguage();
 
+  useBeforeUnload('unload', () => {});
```
TomlDev, line 59: Since you already added this, I think you can do `useBeforeUnload('unload', updateSurveyStorage);` 👌🏽

TomlDev (changes requested):
Very nice and clean solution. 👍🏽 Please have a last look on the comments

TomlDev (approved):

PR #613, opened 2025-03-12
558 - Survey Creator: Analyse why the expiration date does reset with backspace (sometimes)

maierdominik:
Issue: https://github.com/edulution-io/edulution-ui/issues/558

Replace deprecated PR from (https://github.com/edulution-io/edulution-ui/pull/596)

The new shadcn component is derived from https://time.rdsx.dev/

558:
* add the new DateTimePicker component
* fix all the eslint errors
* adapt the color/variant (using the DropdownVariants and Styles)
* use the isDisabled property to disable dates from the past
* add a clear button to the field

Additionally (Fix for Merge of 560 (https://github.com/edulution-io/edulution-ui/pull/607)):
* convert the function getFirstValidDateOfArray into an arrow function


apps/frontend/src/components/ui/DateTimePickerField.tsx
TomlDev, line 1: It must be somehow possible to clear the date
image
I think a boolean that controls it `isClearable` would be good. When a date is entered it should show a button to clear it then.

TomlDev, line 1: Can you please have a look on the AI proposal for avoiding the eslint disable:

> Below is an example of how you could refactor your code to avoid ESLint errors for:
> 
> - **`react/jsx-no-bind`:** by removing inline arrow functions within JSX (moving them to stable callbacks and/or separate child components).  
> - **`@typescript-eslint/no-unsafe-call`:** by doing a proper type check (i.e. verifying the value is actually a `Date`) before calling methods like `.getHours()` or `.getMinutes()`.
> 
> <details>
> <summary>Refactored <code>DateTimePickerField.tsx</code> <strong>(CLICK TO EXPAND)</strong></summary>
> 
> ```tsx
> 'use client';
> 
> import React, { useCallback, useMemo } from 'react';
> import { format } from 'date-fns';
> import { useTranslation } from 'react-i18next';
> import { FieldValues, Path, PathValue, UseFormReturn } from 'react-hook-form';
> import { CalendarIcon } from '@radix-ui/react-icons';
> import { DropdownVariant } from '@libs/ui/types/DropdownVariant';
> import cn from '@libs/common/utils/className';
> import { Button } from '@/components/shared/Button';
> import { Calendar } from '@/components/ui/Calendar';
> import { ScrollArea, ScrollBar } from '@/components/ui/ScrollArea';
> import { Form, FormControl, FormFieldSH, FormItem, FormMessage } from '@/components/ui/Form';
> import { Popover, PopoverContent, PopoverTrigger } from '@/components/ui/Popover';
> 
> /** Utility to safely retrieve hours/minutes. Avoids @typescript-eslint/no-unsafe-call */
> function safeGetHours(dateValue: unknown): number {
>   return dateValue instanceof Date ? dateValue.getHours() : 0;
> }
> function safeGetMinutes(dateValue: unknown): number {
>   return dateValue instanceof Date ? dateValue.getMinutes() : 0;
> }
> 
> /** Separate child component to avoid inline arrow functions in .map for hours */
> interface HourButtonProps {
>   hour: number;
>   currentHour: number;
>   onChangeHour: (hour: number) => void;
>   variant: DropdownVariant;
> }
> function HourButton({ hour, currentHour, onChangeHour, variant }: HourButtonProps) {
>   // stable callback to avoid inline arrow
>   const handleClick = useCallback(() => {
>     onChangeHour(hour);
>   }, [hour, onChangeHour]);
> 
>   return (
>     <Button
>       variant={currentHour === hour ? 'btn-outline' : 'btn-small'}
>       className={cn(
>         'aspect-square max-h-[25px] max-w-[64px] shrink-0 sm:w-full',
>         variant === 'default' && 'bg-background text-foreground',
>         variant === 'dialog' && 'bg-muted text-secondary',
>       )}
>       onClick={handleClick}
>     >
>       {hour}
>     </Button>
>   );
> }
> 
> /** Separate child component to avoid inline arrow functions in .map for minutes */
> interface MinuteButtonProps {
>   minute: number;
>   currentMinute: number;
>   onChangeMinute: (minute: number) => void;
>   variant: DropdownVariant;
> }
> function MinuteButton({ minute, currentMinute, onChangeMinute, variant }: MinuteButtonProps) {
>   const handleClick = useCallback(() => {
>     onChangeMinute(minute);
>   }, [minute, onChangeMinute]);
> 
>   return (
>     <Button
>       variant={currentMinute === minute ? 'btn-outline' : 'btn-small'}
>       className={cn(
>         'aspect-square max-h-[25px] max-w-[64px] shrink-0 sm:w-full',
>         variant === 'default' && 'bg-background text-foreground',
>         variant === 'dialog' && 'bg-muted text-secondary',
>       )}
>       onClick={handleClick}
>     >
>       {minute.toString().padStart(2, '0')}
>     </Button>
>   );
> }
> 
> interface DateTimePickerFieldProps<T extends FieldValues> {
>   form: UseFormReturn<T>;
>   path: Path<T>;
>   translationId?: string;
>   variant?: DropdownVariant;
> }
> 
> const DateTimePickerField = <T extends FieldValues>(props: DateTimePickerFieldProps<T>) => {
>   const { form, path, translationId, variant = 'default' } = props;
>   const { t } = useTranslation();
> 
>   const fieldValue = form.watch(path); // current value of the field
>   const hours = safeGetHours(fieldValue);
>   const minutes = safeGetMinutes(fieldValue);
> 
>   /** stable callback for date selection */
>   const handleDateSelect = useCallback(
>     (date: Date | undefined) => {
>       if (!date) {
>         // Clear the field
>         form.setValue(path, date as PathValue<T, Path<T>>);
>         return;
>       }
> 
>       const currentDate = fieldValue instanceof Date ? fieldValue : new Date();
>       const newDate = new Date(currentDate);
> 
>       newDate.setDate(date.getDate());
>       newDate.setMonth(date.getMonth());
>       newDate.setFullYear(date.getFullYear());
> 
>       form.setValue(path, newDate as PathValue<T, Path<T>>);
>     },
>     [fieldValue, form, path],
>   );
> 
>   /** stable callback for enabling the "disabled dates" behavior on the Calendar */
>   const isDateDisabled = useCallback((date: Date) => date < new Date(), []);
> 
>   /** stable callbacks for setting hours or minutes */
>   const onChangeHour = useCallback(
>     (hour: number) => {
>       const currentDate = fieldValue instanceof Date ? fieldValue : new Date();
>       const newDate = new Date(currentDate);
>       newDate.setHours(hour);
> 
>       form.setValue(path, newDate as PathValue<T, Path<T>>);
>     },
>     [fieldValue, form, path],
>   );
> 
>   const onChangeMinute = useCallback(
>     (minute: number) => {
>       const currentDate = fieldValue instanceof Date ? fieldValue : new Date();
>       const newDate = new Date(currentDate);
>       newDate.setMinutes(minute);
> 
>       form.setValue(path, newDate as PathValue<T, Path<T>>);
>     },
>     [fieldValue, form, path],
>   );
> 
>   /** Memoize the time string for the button's label (avoid re-formatting on every render) */
>   const timeDisplay = useMemo(() => {
>     if (fieldValue instanceof Date) {
>       return format(fieldValue, 'MM/dd/yyyy HH:mm');
>     }
>     return 'MM/DD/YYYY HH:mm';
>   }, [fieldValue]);
> 
>   return (
>     <Form {...form}>
>       <FormFieldSH
>         control={form.control}
>         name={path}
>         render={() => (
>           <FormItem className="flex flex-col">
>             {translationId ? <p className="text-m font-bold text-background">{t(translationId)}</p> : null}
> 
>             <Popover>
>               <PopoverTrigger asChild>
>                 <FormControl>
>                   <Button
>                     variant="btn-outline"
>                     className={cn(
>                       'my-0 h-10 w-fit px-3 py-0 pl-3 text-left font-normal',
>                       !fieldValue && 'text-muted-foreground',
>                     )}
>                   >
>                     {timeDisplay}
>                     <CalendarIcon className="ml-auto h-4 w-4 opacity-50" />
>                   </Button>
>                 </FormControl>
>               </PopoverTrigger>
> 
>               <PopoverContent
>                 className={cn('w-auto p-0', {
>                   'bg-background text-foreground': variant === 'default',
>                   'bg-muted text-secondary': variant === 'dialog',
>                 })}
>               >
>                 <div className="sm:flex">
>                   <Calendar
>                     mode="single"
>                     selected={fieldValue instanceof Date ? fieldValue : undefined}
>                     onSelect={handleDateSelect}
>                     initialFocus
>                     disabled={isDateDisabled}
>                   />
> 
>                   <div>
>                     <div className="m-2 flex h-6 justify-center">
>                       {hours.toString().padStart(2, '0')}:
>                       {minutes.toString().padStart(2, '0')} {t('common.clock')}
>                     </div>
>                     <div className="flex flex-col divide-y sm:h-[300px] sm:flex-row sm:divide-x sm:divide-y-0">
>                       <ScrollArea className="w-64 sm:w-auto">
>                         <div className="flex p-2 sm:flex-col">
>                           {Array.from({ length: 24 }, (_, i) => i)
>                             .reverse()
>                             .map((hour) => (
>                               <HourButton
>                                 key={hour}
>                                 hour={hour}
>                                 currentHour={hours}
>                                 onChangeHour={onChangeHour}
>                                 variant={variant}
>                               />
>                             ))}
>                         </div>
>                       </ScrollArea>
> 
>                       <ScrollArea className="w-64 sm:w-auto">
>                         <div className="flex p-2 sm:flex-col">
>                           {Array.from({ length: 12 }, (_, i) => i * 5).map((minute) => (
>                             <MinuteButton
>                               key={minute}
>                               minute={minute}
>                               currentMinute={minutes}
>                               onChangeMinute={onChangeMinute}
>                               variant={variant}
>                             />
>                           ))}
>                         </div>
>                         <ScrollBar orientation="horizontal" className="sm:hidden" />
>                       </ScrollArea>
>                     </div>
>                   </div>
>                 </div>
>               </PopoverContent>
>             </Popover>
> 
>             <FormMessage />
>           </FormItem>
>         )}
>       />
>     </Form>
>   );
> };
> 
> export default DateTimePickerField;
> ```
> </details>
> 
> ---
> 
> ## Explanation of Key Changes
> 
> 1. **Avoiding `@typescript-eslint/no-unsafe-call`:**  
>    - We check whether `fieldValue` is an instance of `Date` before calling `getHours()` or `getMinutes()`.  
>    - We extracted these calls into small helper functions (`safeGetHours`, `safeGetMinutes`) or used inline checks like `fieldValue instanceof Date`.
> 
> 2. **Avoiding `react/jsx-no-bind`:**  
>    - **Stable Callbacks**  
>      - We define all callbacks (e.g., `handleDateSelect`, `onChangeHour`, `onChangeMinute`, `isDateDisabled`) using `useCallback`. This ensures we do not create a new inline function on every render.  
>    - **Separate Child Components (`HourButton` and `MinuteButton`)**  
>      - Instead of doing `map(...) => <Button onClick={() => handleTimeChange(...)} />`, we pass each item to a separate component.  
>      - Each child component then has its own stable callback (`handleClick`) that calls the prop function with the correct argument. This prevents inline arrow functions in the parent JSX.
> 
> 3. **Memoizing Repeated Computations:**  
>    - We use `useMemo` for the combined date-time display string (`timeDisplay`) to avoid re-formatting on every render.
> 
> By applying the above refactoring, you should eliminate lint errors for **both** `react/jsx-no-bind` **and** `@typescript-eslint/no-unsafe-call`, without having to disable or override these rules in your `.eslintrc`.

TomlDev (changes requested):
Looks like a clean approach. Please have a look at my comments 👍🏽

apps/frontend/src/locales/en/translation.json
```diff
@@ -907,7 +907,7 @@
     "cancel": "Cancel",
     "columns": "Columns",
     "table": "Table",
-    "clock": "clock",
+    "clock": "o'clock",
```
TomlDev, line 910: I think the "common" translation of "clock" is "clock". So you would take a different domain here (e.g. `common.datePicker.clock` or something like this)

maierdominik: it is now stored in:
`form.input.dateTimePicker.timeSlot`

apps/frontend/src/components/ui/DateTimePicker/safeGetHours.ts
```diff
@@ -0,0 +1,15 @@
...
+ * This program is free software: you can redistribute it and/or modify it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+
+const safeGetHours = (dateValue: unknown): number => (dateValue instanceof Date ? dateValue.getHours() : 0);
```
TomlDev, line 13: Should be moved to shared lib `libs/src/common/utils`, create a new folder `Date`

apps/frontend/src/components/ui/DateTimePicker/safeGetMinutes.ts
TomlDev, line 1: Should be moved to shared lib `libs/src/common/utils`, create a new folder `Date`

apps/frontend/src/components/ui/DateTimePicker/safeGetDate.ts
TomlDev, line 1: Should be moved to shared lib `libs/src/common/utils`, create a new folder `Date`

libs/src/common/utils/getFirstValidDateOfArray.ts
TomlDev, line 1: Should be moved to shared lib `libs/src/common/utils`, create a new folder `Date`

TomlDev (changes requested):
It still has English time format and month name:
grafik

apps/frontend/src/components/ui/DateTimePicker/DateTimePickerField.tsx
```diff
@@ -0,0 +1,194 @@
...
+import HourButton from '@/components/ui/DateTimePicker/HourButtons';
+import safeGetDate from '@/components/ui/DateTimePicker/safeGetDate';
+
+interface DateTimePickerFieldProps<T extends FieldValues> {
+  form: UseFormReturn<T>;
+  path: Path<T>;
+  translationId?: string;
+  variant?: DropdownVariant;
+}
+
```
TomlDev, line 40: The delete button should not be visible when there is no date set

TomlDev, line 40: Currently when you clear the date and press save, it still takes the old date.

apps/frontend/src/components/ui/DateTimePicker/DateTimePickerField.tsx
```diff
@@ -42,12 +44,14 @@ const DateTimePickerField = <T extends FieldValues>(props: DateTimePickerFieldPr
   const { form, path, translationId, variant = 'default' } = props;
   const { t } = useTranslation();
 
+  const language = useLanguage();
```
TomlDev, line 47: const { language } = useLanguage();

apps/frontend/src/components/ui/DateTimePicker/DateTimePickerField.tsx
```diff
@@ -85,11 +89,22 @@ const DateTimePickerField = <T extends FieldValues>(props: DateTimePickerFieldPr
 
   const isDateDisabled = useCallback((date: Date) => date < new Date(), []);
 
+  const locale = useMemo(() => {
+    if (language?.language === 'de') return de;
+    return enUS;
+  }, [language]);
```
TomlDev, line 95: > In this code snippet, **`useMemo`** is being used twice:
> 
> ```ts
> const locale = useMemo(() => {
>   if (language?.language === 'de') return de;
>   return enUS;
> }, [language]);
> 
> const timeDisplay = useMemo(() => {
>   if (language.language === 'en') {
>     if ((fieldValue as unknown) instanceof Date) {
>       return format(fieldValue, 'MM/dd/yyyy HH:mm', { locale });
>     }
>     return 'MM/DD/YYYY HH:mm';
>   }
>   if ((fieldValue as unknown) instanceof Date) {
>     return format(fieldValue, 'dd/MM/yyyy HH:mm', { locale });
>   }
>   return 'DD/MM/YYYY HH:mm';
> }, [fieldValue]);
> ```
> 
> ### Why `useMemo` at all?
> 
> - **`locale`**: It’s a simple conditional (`if (...) return de; else return enUS;`). Using `useMemo` here avoids recreating a new locale object on every render – though it’s not a huge expense to begin with. That said, if you pass `locale` as a prop to a child component, memoizing it can help prevent unnecessary child re-renders due to a changed object reference.
> - **`timeDisplay`**: It runs a small `format` function call from `date-fns`, which also isn’t typically super expensive. Still, if this code runs often (e.g. multiple items in a list), you might want to memoize it to avoid repeated re-formatting. Or, if you pass `timeDisplay` into a child component as a prop, memoizing keeps that prop stable as long as the underlying values haven’t changed.
> 
> In short, **`useMemo`** can be beneficial if:
> 1. The calculation inside is non-trivial, or  
> 2. You pass its result as a prop into child components you want to keep memoized.  
> 
> Otherwise, it might be a premature optimization. It won’t *break* anything, but you have extra complexity for potentially marginal performance benefit.
> 
> ---
> 
> ### The Bigger Issue: Dependencies
> 
> Right now, your dependencies are:
> 
> ```ts
> [language] // for locale
> [fieldValue] // for timeDisplay
> ```
> 
> But in `timeDisplay` you also read from `language.language` and from `locale`. So you likely need:
> 
> ```ts
> const timeDisplay = useMemo(() => {
>   // uses language.language and locale
> }, [fieldValue, language, locale]);
> ```
> 
> Otherwise, if the user changes their language, `timeDisplay` won’t recalculate (because `useMemo` thinks only `fieldValue` changed). Or if `locale` changes, `timeDisplay` also won’t recalc. That can lead to stale UI.
> 
> Similarly, for `locale`:
> ```ts
> const locale = useMemo(() => {
>   if (language?.language === 'de') return de;
>   return enUS;
> }, [language]);
> ```
> That’s okay, but be sure that `language` is stable (if `language` is an object that changes identity often, you might still get rerenders). Typically we’d do `[language.language]` or something more specific if that’s all that changes.
> 
> ---
> 
> ### Summarizing Recommendations
> 
> 1. **Either remove `useMemo` entirely** if the cost of creating `locale`/formatting a date is negligible. Often a direct inline calculation is just fine – especially if it’s only in a single small component.
> 2. **Or keep `useMemo`, but fix the dependencies**. For `timeDisplay`, you should include `[fieldValue, language, locale]` to ensure it recalculates when the user changes language or if `locale` changes.
> 3. **Check if you actually need the memoization**. If you’re not passing these values into child components or performing heavy computations, you can simplify the code and skip `useMemo`.  
> 
> In many real-world React apps, `useMemo` is most useful when rendering large lists, expensive computations, or if you have memoized child components that depend on stable references. Otherwise, it may be an unnecessary micro-optimization.

apps/frontend/src/components/ui/DateTimePicker/DateTimePickerField.tsx
```diff
@@ -85,11 +89,22 @@ const DateTimePickerField = <T extends FieldValues>(props: DateTimePickerFieldPr
...
+        return format(fieldValue, 'MM/dd/yyyy HH:mm', { locale });
+      }
+      return 'MM/DD/YYYY HH:mm';
+    }
     if ((fieldValue as unknown) instanceof Date) {
-      return format(fieldValue, 'MM/dd/yyyy HH:mm');
+      return format(fieldValue, 'dd/MM/yyyy HH:mm', { locale });
     }
-    return 'MM/DD/YYYY HH:mm';
+    return 'DD/MM/YYYY HH:mm';
```
TomlDev, line 107: Isnt it possible to use something like `.toLocaleString('de')` from Date?

libs/src/survey/types/api/survey.dto.ts
```diff
@@ -29,7 +29,7 @@ interface SurveyDto {
   participatedAttendees: AttendeeDto[];
   answers: string[];
   createdAt?: Date;
-  expires?: Date;
+  expires?: Date | null;
```
TomlDev, line 32: Should it still be optional then?

TomlDev (changes requested):
Currently when you save a survey and select a date for the first time you can select the same day. But after you selected any other day you cannot select the current day again.
I think it should be possible to create a survey that expires the same day (later)

TomlDev (changes requested):
Can you please style the component like our input fields:
grafik

TomlDev (approved):

libs/src/ui/types/DropdownVariant.ts
```diff
@@ -0,0 +1,13 @@
...
+ * This program is free software: you can redistribute it and/or modify it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+
+export type DropdownVariant = 'dialog' | 'default';
```
michidold, line 13: Use default export if there is only one type

libs/src/common/utils/sortDate.ts
```diff
@@ -10,9 +10,9 @@
  * You should have received a copy of the GNU Affero General Public License along with this program. If not, see <https://www.gnu.org/licenses/>.
  */
 
-function sortDate(a?: Date, b?: Date): number {
-  if (!b) return !a ? 0 : 1;
-  if (!a) return -1;
+function sortDate(a: Date | null | undefined, b: Date | null | undefined): number {
```
michidold, line 13: Use a arrow function. And can this also be moved to the Date subfolder?

apps/frontend/src/pages/BulletinBoard/BulletinBoardEditorial/CreateOrUpdateBulletinDialogBody.tsx
michidold, line 1: Here are two issues:
1. form resolver throws an error:
image

The schema could be adjusted to:
```typescript
import { z } from 'zod';
import { TFunction } from 'i18next';

const getBulletinFormSchema = (t: TFunction<'translation', undefined>) =>
  z
    .object({
 ...

      isVisibleStartDate: z.date().nullable().optional(),
      isVisibleEndDate: z.date().nullable().optional(),
    })
   ...

export default getBulletinFormSchema;
```

2. On reload the bulletin,  the inserted date is formatted wrong:
image

michidold (changes requested):
Found some minor issues. Please have a look.

michidold (approved):
:+1:

PR #653, opened 2025-03-24
[Classmanagement, Filesharing] Move dialog navigating in empty folder gives errors

Maddi02:
Replace direct basename usage with getFileNameFromPath utility for consistent and cleaner path extraction. Added the utility import to ensure proper functionality.


Maddi02:
In the issue, it was initially reported that navigating into an empty folder in the Move or Share dialog caused an error. However, the real problem turned out to be the difference between the basename and the filename. In the Move/Share File dialog, the path is constructed using the basename, while in the File Sharing table it relies on the filename. For instance, a folder named `20` was identified as `020` instead of `20.` Therefore, the only required fix is to adjust the navigation in the Share/Move dialog. Refer to the code changes for more details.

TomlDev (commented):
Do you know why the folder's name is 020 and sometimes displayed 20?
grafik

Maddi02:
Yes, the reason is Filename and basename are different in the response provided by WebDAV.

```
   {
        "basename": "20",
        "etag": "f09f7662d4230725ccdb552f19012dd3",
        "filename": "/webdav///teachers/agy-netzint-teacher/transfer/020",
        "lastmod": "Fri, 05 Jul 2024 13:42:35 GMT",
        "type": "COLLECTION"
    },
```

TomlDev (approved):

PR #670, opened 2025-03-28
Bulletin Board: Always return latest bulletin

TomlDev:
Instead of the newest created one I would suggest to sort by `updatedAt` 👍🏽


michidold (approved):
:+1:

PR #685, opened 2025-04-02
673 - Login via edu APP

michidold:


apps/frontend/src/pages/LoginPage/LoginPage.tsx
TomlDev, line 1: When the request from the APP is sent and the frontend receives the values it somehow recreates the QR code.
If you don't send the correct credentials and then scan the "new" QR code it seems to be an invalid one (without a working SSE connection).

TomlDev, line 1: Also when you enter the correct credentials it somehow recreates the QR code in the moment before you are logged in:

https://github.com/user-attachments/assets/9f299671-8aa7-43a6-93a6-310b2508589f

You can see the QR code changing right before the user is logged in

TomlDev, line 1: When you send a wrong TOTP or none **when** TOTP is necessary you don't get an error.
I think for the App it will be necessary to respond with a meaningful message like 'TOTP invalid'

TomlDev (changes requested):
Clean code, just some small considerations

apps/frontend/src/pages/LoginPage/LoginPage.tsx
```diff
@@ -129,6 +144,60 @@ const LoginPage: React.FC = () => {
...
+      try {
+        const { username, password, totpValue } = JSON.parse(atob(e.data)) as LoginQrSseDto;
+
+        form.setValue('username', username);
+        form.setValue('password', password);
+        form.setValue('totpValue', totpValue || '');
+
+        void form.handleSubmit(onSubmit)();
+      } catch (error) {
+        console.error('JSON parse error:', error);
```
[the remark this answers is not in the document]
michidold: The useform hook has its own error handler due to its resolver middleware.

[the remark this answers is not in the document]
TomlDev: Then I would move line 166 outside of the try catch

apps/frontend/src/pages/LoginPage/LoginPage.tsx
michidold: Unfortunately, there is no way to inform the App that login failed. But I added a toast.error, if qr login is active and login fails

TomlDev (approved):
👌🏽 Nice!

PR #700, opened 2025-04-07
695 Use a variable where window.location.origin is used

TomlDev:


libs/src/common/constants/eduApiBaseUrl.ts
```diff
@@ -0,0 +1,15 @@
...
+ * This program is free software: you can redistribute it and/or modify it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+
+const EDU_API_BASE_URL = window.location.origin;
```
michidold, line 13: Think its more like EDU_BASE_URL since it is the frontend url.

michidold (approved):
Looks good, only one thing. Could we replace EDU_API_BASE_URL with EDU_BASE_URL?

PR #703, opened 2025-04-08
572 Scrollable Table should not overflow page content

TomlDev:
- [x] Set table to fixed layout
- [x] Overwrite table header column width if size is defined and it differs from the default value
- [x] Add small margin to sorting arrows for better mobile support


apps/frontend/src/components/ui/Table/ScrollableTable.tsx
michidold, line 1: Don't know if its related to theses changes specifically, but in AppConfigPage every table has that cut of left border in the tables' search field. We had accessed that by adding a small padding `pl-1` maybe that is getting lost during rework.

Bildschirmfoto vom 2025-04-08 15-18-01

michidold (approved):
Found only one small issue.

##FFTR

PR #709, opened 2025-04-09
[Filesharing] Dropdown for column sections has incorrect background color

Maddi02:


Maddi02:
I've fixed the background color issue as shown in the screenshot below:

image

However, since there is only a single option available, I think displaying a dropdown in this case is unnecessary. Therefore, I introduced a condition that ensures the dropdown for the columns is only shown when there are at least two elements available in the dropdown list.

So we ended up with this 

image

TomlDev (approved):
Good job 👌🏽

michidold (approved):
:+1:

PR #711, opened 2025-04-09
Move Content Dialog: Remove unused space

TomlDev:
After:
grafik

`showSelectedCount` is only `false` for `<MoveContentDialogBody />`


apps/frontend/src/components/ui/Table/ScrollableTable.tsx
michidold, line 1: Not related to this task, but why is your column selector missing?
grafik

michidold (approved):
:+1:

apps/frontend/src/components/ui/Table/ScrollableTable.tsx
TomlDev: I made the changes on top of MHs branch: https://github.com/edulution-io/edulution-ui/pull/709#issuecomment-2788518116
But afterwards used `dev` as base 👍🏽

PR #715, opened 2025-04-16
509-Update outdated deps and fix router issues

michidold:
- Run npm update
- Fix router issues
==> react-router with version >= 6.29 won't redirect at logout anymore. Further it throws a error when routes change dynamically, like on using isSuperAdmin to change routes.

### DIFF
--- a/apps/frontend/src/hooks/useLogout.tsx
+++ b/apps/frontend/src/hooks/useLogout.tsx
@@ -14,6 +14,7 @@ import { useCallback } from 'react';
 import useUserStore from '@/store/UserStore/UserStore';
 import { useAuth } from 'react-oidc-context';
 import cleanAllStores from '@/store/utils/cleanAllStores';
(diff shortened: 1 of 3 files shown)


TomlDev (changes requested):
Please have a look, it didn't redirect me after the login when I tried to access an existing route:

https://github.com/user-attachments/assets/39de1ccf-2930-4fbe-8bd8-cf7ae8fc0481

michidold:
> Please have a look, it didn't redirect me after the login when I tried to access an existing route:
> Peek.2025-04-17.13-28.mp4

Seems to be corrupt since the big layout rework. Actually, the appconfig is not loaded yet when login is finished, so generating theses routes is not finished ether. I fixed it, by waiting to have the appconfigs loaded before redirect.

Maddi02 (approved):
👍

PR #718, opened 2025-04-17
374 Unify styling of buttons and actions for close and cancel

michidold:
- Add DialogFooterButtons component
- Replaced or added dialog buttons


Maddi02 (approved):
Nice solution 👍

PR #723, opened 2025-04-22
Fix styling 20250422

michidold:
- [x] Moved dashboard description right under the title
- [x] Reworked styling of classmgmt cards on all 4 pages (enrol, password, projects and lessons)
- [x] Removed school prefixes from cards

### DIFF
--- a/apps/frontend/src/locales/de/translation.json
+++ b/apps/frontend/src/locales/de/translation.json
@@ -2,7 +2,7 @@
   "welcome": "Willkommen in der Anwendung",
   "changeLang": "Sprache ändern",
   "heading": "Herzlich Willkommen {{givenName}} {{familyName}}",
-  "content": "Hier ist ihr persönliches {{ applicationName }} Dashboard.",
+  "content": "Hier ist ihr persönliches {{ applicationName }} Dashboard. Dem Ort an dem all ihre Anwendungen für den digitalen Unterricht gesammelt sind.",
   "closeEditingWindow": "Will du wirklich das Bearbeitungsfenster schließen?",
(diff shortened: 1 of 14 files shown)


apps/frontend/src/locales/de/translation.json
```diff
@@ -932,7 +932,7 @@
     "forgot_password": "Passwort vergessen?",
     "remember_me": "Merken",
     "enterMultiFactorCode": "Zwei-Faktor-Authentisierung",
-    "loginWithApp": "Login mit QR",
+    "loginWithApp": "Anmelden mit QR-Login",
```
Maddi02, line 935: QR Login? Or should we call it "with App Login"?  

But I think we should agree on one name. To ensure consistency of name across app and web.

apps/frontend/src/locales/en/translation.json
```diff
@@ -930,7 +930,7 @@
     "forgot_password": "Forgot password?",
     "remember_me": "Remember me",
     "enterMultiFactorCode": "Two-Factor Authentication",
-    "loginWithApp": "Login with QR",
+    "loginWithApp": "Login with QR-Login",
```
Maddi02, line 933: Same as above

Maddi02 (changes requested):
image


Looks good. However, in the class management page, the selected group is invisible. Could you have a look?

apps/frontend/src/locales/de/translation.json
michidold: That's on purpose

Maddi02 (approved):
👍

PR #726, opened 2025-04-24
Fix save appconfig throws error file not found

michidold:
Added a check, if a file exists before deletion


Maddi02 (approved):
👍 Works well :)

PR #728, opened 2025-04-24
Feature - Add tldraw to default app config

michidold:
- Removed excalidraw lib and assets
- Reworked apps to include only one whiteboard component
- Renamed whiteboardcollaboration to just whiteboard


Maddi02 (approved):
Looks fine. However, I keep getting this error


image

PR #743, opened 2025-04-28
743 Filesharing: The file preview is loaded regardless of whether the file type could be displayed or not.

Maddi02:


libs/src/filesharing/types/previewableExtensions.ts
TomlDev, line 1: This is not a type, its a util function

libs/src/filesharing/types/videoExtensions.ts
TomlDev, line 1: This should be a const.

libs/src/filesharing/types/audioExtensions.ts
TomlDev, line 1: This should be a const.

libs/src/filesharing/utils/isFileValid.ts
TomlDev, line 1: What about renaming this to `isValidFileToPreview` to make it more clear what makes this a "valid" file?
And could this not be reused instead of the `isPreviewableExtension`?

apps/frontend/src/components/ui/MediaComponent.tsx
```diff
@@ -40,10 +40,7 @@ const VideoComponent: React.FC<VideoComponentProps> = ({
   height = '100%',
   style = {},
 }) => (
-  <div
-    className="video-player"
-    style={{ position: 'relative', paddingTop: '56.25%', ...style }}
-  >
+  <div style={{ position: 'relative', paddingTop: '56.25%', ...style }}>
```
TomlDev, line 43: Do you know why the `paddingTop` is necessary? If it should be centered, isn't there a better way?

TomlDev (changes requested):
Please have a look :+1

apps/frontend/src/components/ui/MediaComponent.tsx
Maddi02: It isnt necessary, but I now have an idea of why it is in place.

TomlDev (approved):

michidold (changes requested):
Could not play webm files. Maybe we should test all the listed file types, or shorten the list.

libs/src/filesharing/utils/isMediaExtension.ts
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+import VIDEO_EXTENSIONS from '@libs/filesharing/types/videoExtensions';
+import VideoExtensionType from '@libs/filesharing/types/videoExtensionType';
+import AudioExtensionsType from '@libs/filesharing/types/audioExtensionsType';
+import AUDIO_EXTENSIONS from '@libs/filesharing/types/audioExtensions';
+
+const isMediaExtension = (extension: string | undefined): boolean =>
+  Object.values(VIDEO_EXTENSIONS).includes(extension as VideoExtensionType) ||
+  Object.values(AUDIO_EXTENSIONS).includes(extension as AudioExtensionsType);
```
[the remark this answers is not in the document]
Maddi02: Yes. During testing, I removed the file types that did not work.

michidold (approved):
:+1:

PR #762, opened 2025-04-30
Add SSO Keycloak login with hidden frame

michidold:
- Added a hook to login to Keycloak via a hidden iframe to get the SSO cookie
- Add silent logout
- Renamed auth-endpoints file to auth-paths corresponding to the const
- Add new Vite env for Keycloak url

To test it locally you need to ether deploy a Keycloak instance or simpler map Keycloak port locally:

```zsh
ssh username@server -L 8080:localhost:8080
```

Therefore the keycloak instance on the server must have mapped port 8080 to localhost. Default its just inside the docker network available. Please refer to me to setup the test environment

After login, there should be KEYCLOAK_IDENTITY and KEYCLOAK_SESSION cookies available:

grafik


apps/frontend/src/pages/LoginPage/useSilentKeycloakLogin.ts
```diff
@@ -0,0 +1,114 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+
+/* eslint-disable @typescript-eslint/no-use-before-define */
+import { useCallback } from 'react';
+import Keycloak from 'keycloak-js';
+import EDU_BASE_URL from '@libs/common/constants/eduApiBaseUrl';
+import AUTH_PATHS from '@libs/auth/constants/auth-paths';
```
TomlDev, line 17: Are these really paths? E.g.
```
    realm: AUTH_PATHS.KEYCLOAK_REALM,
    clientId: AUTH_PATHS.KEYCLOAK_CLIENT_ID,

File: apps/frontend/src/pages/LoginPage/useSilentKeycloakLogin.ts
Comment (TomlDev) [inline apps/frontend/src/pages/LoginPage/useSilentKeycloakLogin.ts:1]: Please have a look on the AI proposed refactoring. Would it be possible to extract the iframe creation logic?

> Below is one way to cleanly split concerns, reduce nesting and duplication, and make the hook easier to test and extend:
> 
> ```ts
> import { useCallback, useMemo } from 'react';
> import Keycloak, { KeycloakConfig } from 'keycloak-js';
> import EDU_BASE_URL from '@libs/common/constants/eduApiBaseUrl';
> import AUTH_PATHS from '@libs/auth/constants/auth-paths';
> 
> interface UseSilentLoginWithPasswordReturn {
>   silentLogin: (username: string, password: string) => Promise<void>;
>   silentLogout: () => Promise<void>;
> }
> 
> // --- Helpers -------------------------------------------------------------
> 
> /**
>  * Creates a hidden iframe, sets src, waits for load, then returns it.
>  */
> function createHiddenIframe(src: string): Promise<HTMLIFrameElement> {
>   return new Promise(resolve => {
>     const iframe = document.createElement('iframe');
>     iframe.style.display = 'none';
>     iframe.src = src;
>     iframe.onload = () => resolve(iframe);
>     document.body.appendChild(iframe);
>   });
> }
> 
> /**
>  * Listens for a single postMessage from same-origin, then cleans up.
>  */
> function waitForMessage(): Promise<string> {
>   return new Promise(resolve => {
>     const handler = (ev: MessageEvent) => {
>       if (ev.origin !== window.location.origin) return;
>       window.removeEventListener('message', handler);
>       resolve(typeof ev.data === 'string' ? ev.data : window.location.href);
>     };
>     window.addEventListener('message', handler);
>   });
> }
> 
> // --- Hook ---------------------------------------------------------------
> 
> const useSilentLoginWithPassword = (): UseSilentLoginWithPasswordReturn => {
>   const keycloakConfig: KeycloakConfig = useMemo(() => ({
>     url: new URL('auth', EDU_BASE_URL).href,
>     realm: AUTH_PATHS.KEYCLOAK_REALM,
>     clientId: AUTH_PATHS.KEYCLOAK_CLIENT_ID,
>   }), []);
> 
>   // instantiate just once
>   const keycloak = useMemo(() => new Keycloak(keycloakConfig), [keycloakConfig]);
> 
>   const redirectUri = useMemo(
>     () => new URL('silent-check-sso.html', EDU_BASE_URL).href,
>     []
>   );
> 
>   const silentLogin = useCallback(
>     async (username: string, password: string) => {
>       // 1) initialize SSO iframe check
>       await keycloak.init({
>         onLoad: 'check-sso',
>         silentCheckSsoRedirectUri: redirectUri,
>         pkceMethod: 'S256',
>         checkLoginIframe: false,
>       });
> 
>       // 2) build the login flow URL
>       const authUrl = await keycloak.createLoginUrl({
>         prompt: 'login',
>         redirectUri,
>       });
> 
>       // 3) fire off the iframe + wait for form to load
>       const iframe = await createHiddenIframe(authUrl);
> 
>       // 4) submit the username/password
>       const doc = iframe.contentDocument;
>       const form = doc?.querySelector<HTMLFormElement>('form');
>       const userInput = doc?.querySelector<HTMLInputElement>('input[name="username"]');
>       const passInput = doc?.querySelector<HTMLInputElement>('input[name="password"]');
>       if (!form || !userInput || !passInput) {
>         iframe.remove();
>         throw new Error('Login form not found');
>       }
>       userInput.value = username;
>       passInput.value = password;
>       form.submit();
> 
>       // 5) wait for the redirect message, then tear down
>       await waitForMessage();
>       iframe.remove();
>       console.info('Login successful');
>     },
>     [keycloak, redirectUri],
>   );
> 
>   const silentLogout = useCallback(async () => {
>     const logoutUrl = `${keycloakConfig.url}/realms/${keycloakConfig.realm}/protocol/openid-connect/logout`
>       + `?post_logout_redirect_uri=${EDU_BASE_URL}`
>       + `&client_id=${keycloakConfig.clientId}`;
> 
>     const iframe = await createHiddenIframe(logoutUrl);
>     iframe.remove();
>     console.info('Logout successful');
>   }, [keycloakConfig]);
> 
>   return { silentLogin, silentLogout };
> };
> 
> export default useSilentLoginWithPassword;
> ```
> 
> ---
> 
> ### What changed and why
> 
> 1. **`createHiddenIframe`**
> 
>    * Encapsulates DOM creation + load-promise.
>    * Avoids repeating that in both login & logout.
> 
> 2. **`waitForMessage`**
> 
>    * Isolates the `postMessage` listener, teardown, and promise logic.
> 
> 3. **`useMemo` for configs + Keycloak instance**
> 
>    * Ensures you only ever create one Keycloak client.
>    * Keeps your dependency arrays minimal.
> 
> 4. **Flattened `silentLogin`**
> 
>    * No more nested `new Promise` blocks.
>    * Sequential steps are clear: init → get URL → iframe → form submit → wait for message.
> 
> 5. **Error handling**
> 
>    * Throws if the form or inputs aren’t found, rather than hanging forever.
> 
> 6. **Consistent tear-down**
> 
>    * Every path removes its iframe and event listener.
> 
> This refactor makes each piece independently testable, easier to read, and removes duplication between login and logout.
>

**Review by TomlDev — CHANGES_REQUESTED** [review]
Could not fully test it

File: apps/frontend/src/pages/LoginPage/useSilentKeycloakLogin.ts [hunk 2]
```diff
@@ -0,0 +1,125 @@
...
+          if (ev.origin !== window.location.origin) return;
+          const href = typeof ev.data === 'string' ? ev.data : window.location.href;
+          const params = new URL(href).searchParams;
+          const hashParams = new URL(href).hash.slice(1);
+          const code = params.get('code') ?? new URLSearchParams(hashParams).get('code');
+
+          if (!code) {
+            cleanup();
+          }
+          cleanup();
```
TomlDev, line 65: The conditional is unnecessary  FFTR after solving

TomlDev (approved):

Issue #785, opened 2025-05-05
Survey - backend limiter - check the visualization of the survey result table when using backend limiter


maierdominik:
![Image](https://github.com/user-attachments/assets/5c91a6c7-1cb9-410a-bac6-05032368b708)

PR #788, opened 2025-05-05
785 - Survey - backend limiter - check the visualization of the survey result table when using backend limiter

maierdominik:
Issue: https://github.com/edulution-io/edulution-ui/issues/785

785:
 * use the choices.title as valueName and titleName because surveyParticipationModel.getData() only returns the name (id) or the title (value) i enforce the title as answer

We would have to update the Surveys and Templates by adding following two lines:
image


michidold (approved):
:+1:

TomlDev (approved):

PR #789, opened 2025-05-06
TLDraw Collabo: Check token in websocket gateway

TomlDev:
Unfortunately it's not possible to inject our AuthGuard, nor to use the Cookie for authentication.


michidold (approved):
:+1:

PR #794, opened 2025-05-06
Implement move copy with bulk processing support

Maddi02:
This PR introduces a solution for situations in which a user needs to move or copy multiple files simultaneously. This is because WebDAV cannot process multiple MOVE or COPY requests concurrently. This PR queues each operation to ensure the workflow succeeds, even if one operation fails.

Currently, each MOVE, COPY or DELETE action is sent as a separate request, so moving or copying N files results in N calls. This PR replaces that pattern with a single API call carrying an array of move or copy commands.


michidold (changes requested):

apps/api/src/filesharing/consumers/copyFile.consumer.ts
```diff
@@ -0,0 +1,79 @@
...
+import { WorkerHost } from '@nestjs/bullmq';
+import { Job } from 'bullmq';
+import { Injectable } from '@nestjs/common';
+import * as path from 'path';
+
+import FilesharingProgressDto from '@libs/filesharing/types/filesharingProgressDto';
+import SSE_MESSAGE_TYPE from '@libs/common/constants/sseMessageType';
+import FileOperationQueueJobData from '@libs/queue/constants/fileOperationQueueJobData';
+import getStudentNameFromPath from '@libs/filesharing/utils/getStudentNameFromPath';
+import makeUniqueName from '@libs/filesharing/utils/makeUniqueName';
```
TomlDev, line 22: create instead of make?
even better depending on how it creates: generate

apps/api/src/filesharing/filesharing.controller.ts
```diff
@@ -110,17 +111,15 @@ class FilesharingController {
 
   @Patch()
   async moveOrRenameResource(
-    @Body()
-    body: {
-      path: string;
-      newPath: string;
-    },
+    @Body() pathChangeOrCreateDto: PathChangeOrCreateDto[],
```
TomlDev, line 114: It should be plural since its an array

apps/api/src/filesharing/filesharing.controller.ts
```diff
@@ -110,17 +111,15 @@ class FilesharingController {
...
-    const originFull = `${this.baseurl}${path}`;
-    const newFull = `${this.baseurl}${newPath}`;
-    return this.webdavService.moveOrRenameResource(username, originFull, newFull);
+    if (pathChangeOrCreateDto.length === 1) {
+      const originFull = `${this.baseurl}${pathChangeOrCreateDto[0].path}`;
+      const newFull = `${this.baseurl}${pathChangeOrCreateDto[0].newPath}`;
+      return this.webdavService.moveOrRenameResource(username, originFull, newFull);
+    }
+    return this.filesharingService.moveOrRenameResource(username, pathChangeOrCreateDto);
   }
```
TomlDev, line 123: I dont understand why this makes sense. If it's only 1 path, we use the webdav service and if there are multiple paths we use the filesharing service? But both functions have the same name. This looks wrong.

apps/api/src/filesharing/filesharing.controller.ts
```diff
@@ -167,6 +166,11 @@ class FilesharingController {
     return this.filesharingService.duplicateFile(username, duplicateFileRequestDto);
   }
 
+  @Post(FileSharingApiEndpoints.COPY)
+  async copyFile(@Body() pathChangeOrCreateDto: PathChangeOrCreateDto[], @GetCurrentUsername() username: string) {
+    return this.filesharingService.copyFileFolder(username, pathChangeOrCreateDto);
```
TomlDev, line 171: What is a fileFolder?

apps/api/src/filesharing/filesharing.service.ts
```diff
@@ -73,6 +90,22 @@ export default class FilesharingService {
     );
   }
 
+  async moveOrRenameResource(username: string, pathChangeOrCreateDto: PathChangeOrCreateProps[]) {
```
TomlDev, line 93: The functions name must be plural since you pass an array. Same for the parameter

apps/frontend/src/hooks/useNotifications.ts
```diff
@@ -144,13 +144,15 @@ const useNotifications = () => {
...
-    const handleDelete = (e: MessageEvent<string>) => {
-      void rawDelete(e);
+    const handleFileOperations = (e: MessageEvent<string>) => {
+      void rawFileOperations(e);
     };
 
-    eventSource.addEventListener(SSE_MESSAGE_TYPE.FILESHARING_DELETE_FILES, handleDelete, { signal });
+    eventSource.addEventListener(SSE_MESSAGE_TYPE.FILESHARING_DELETE_FILES, handleFileOperations, { signal });
+    eventSource.addEventListener(SSE_MESSAGE_TYPE.FILESHARING_MOVE_OR_RENAME_FILES, handleFileOperations, { signal });
+    eventSource.addEventListener(SSE_MESSAGE_TYPE.FILESHARING_COPY_FILES, handleFileOperations, { signal });
```
TomlDev, line 155: What about this?
```typescript
useEffect(() => {
...

  // this is the one "real" callback we need
  const fileOperationHandler = handleFilesharingProgress(setFileOperationProgress);

  // attach the same handler to all three event types
  [
    SSE_MESSAGE_TYPE.FILESHARING_DELETE_FILES,
    SSE_MESSAGE_TYPE.FILESHARING_MOVE_OR_RENAME_FILES,
    SSE_MESSAGE_TYPE.FILESHARING_COPY_FILES,
  ].forEach((type) => {
    eventSource.addEventListener(type, fileOperationHandler, { signal });
  });

...

File: apps/frontend/src/pages/FileSharing/Dialog/handleFileAction/handleBulkFileOperations.ts [hunk 7]
```diff
@@ -48,8 +37,14 @@ const handleBulkFileOperations = async (
...
-    await sendBatchRequests(itemsToProcess, endpoint, httpMethod);
+  } else if (action === FileActionType.MOVE_FILE_FOLDER) {
+    await eduApi[httpMethod](endpoint, itemsToProcess);
+    setFileOperationResult(undefined, t('fileOperationSuccessful'), 200);
+  } else if (action === FileActionType.RENAME_FILE_FOLDER) {
+    await eduApi[httpMethod](endpoint, itemsToProcess);
+    setFileOperationResult(true, t('fileOperationSuccessful'), 200);
+  } else if (action === FileActionType.COPY_FILE_FOLDER) {
+    await eduApi[httpMethod](endpoint, itemsToProcess);
     setFileOperationResult(true, t('fileOperationSuccessful'), 200);
```
TomlDev, line 48: Instead of `200` you could use `HttpStatus.OK`

apps/frontend/src/pages/FileSharing/Dialog/handleFileAction/handleBulkFileOperations.ts
TomlDev, line 1: Please check out this:

> 
> * **Collapsing** all the “do the API call” branches into a single switch/if
> * **Extracting** the one‐off delete logic
> * **Unifying** the “success” flag and status/message outside of the branching
> * **Throwing** (or handling) an error for unsupported actions
> 
> ```ts
> /*
>  * LICENSE
>  *  …
>  */
> 
> import { HttpMethods } from '@libs/common/types/http-methods'
> import getPathWithoutWebdav from '@libs/filesharing/utils/getPathWithoutWebdav'
> import eduApi from '@/api/eduApi'
> import FileActionType from '@libs/filesharing/types/fileActionType'
> import PathChangeOrCreateDto from '@libs/filesharing/types/pathChangeOrCreateProps'
> import buildApiDeletePathUrl from '@libs/filesharing/utils/buildApiDeletePathUrl'
> import DeleteTargetType from '@libs/filesharing/types/deleteTargetType'
> import { t } from 'i18next'
> 
> async function handleDeleteItems(
>   paths: PathChangeOrCreateDto[],
>   endpoint: string
> ): Promise<void> {
>   const cleanPaths = paths.map(p => getPathWithoutWebdav(p.path))
>   const url = buildApiDeletePathUrl(endpoint, DeleteTargetType.FILE_SERVER)
>   await eduApi.delete(url, { data: { paths: cleanPaths } })
> }
> 
> /**
>  * Perform a bulk file operation (delete, move, rename, copy) and then
>  * call back with (successFlag, message, statusCode)
>  */
> export default async function handleBulkFileOperations(
>   action: FileActionType,
>   endpoint: string,
>   httpMethod: HttpMethods,
>   items: PathChangeOrCreateDto[],
>   setResult: (success: boolean | undefined, message: string, statusCode: number) => void,
> ) {
>   try {
>     switch (action) {
>       case FileActionType.DELETE_FILE_FOLDER:
>         await handleDeleteItems(items, endpoint)
>         break
> 
>       case FileActionType.MOVE_FILE_FOLDER:
>       case FileActionType.RENAME_FILE_FOLDER:
>       case FileActionType.COPY_FILE_FOLDER:
>         // MOVE, RENAME and COPY all use the same HTTP call
>         await eduApi[httpMethod](endpoint, items)
>         break
> 
>       default:
>         throw new Error(`Unsupported action: ${action}`)
>     }
> 
>     // only rename & copy explicitly count as “true” success  
>     const success = [
>       FileActionType.RENAME_FILE_FOLDER,
>       FileActionType.COPY_FILE_FOLDER,
>     ].includes(action)
> 
>     setResult(success, t('fileOperationSuccessful'), 200)
>   } catch (err: any) {
>     // you might want more sophisticated error‐parsing here
>     const msg = err?.message ?? t('fileOperationFailed')
>     const code = err?.statusCode ?? 500
>     setResult(false, msg, code)
>   }
> }
> ```
> 
> ### What changed
> 
> 1. **Single `switch`** instead of repetitive `if/else` chains
> 2. **Delete logic** is pulled into its own `handleDeleteItems` (with a clearer variable name)
> 3. **Unified API call** for move/rename/copy—no need to write it three times
> 4. **Post‐call result** is applied once, using a small lookup to decide whether `success` is `true` or `undefined`
> 5. **Error handling** around the whole thing, so any thrown (or network) error will surface as a “failed” result
> 
> This version is shorter, easier to follow, and makes it trivial to add new actions later (just another case in the switch).
>

libs/src/filesharing/types/fileActionType.ts
```diff
@@ -17,6 +17,7 @@ enum FileActionType {
   DELETE_FILE_FOLDER = 'deleteFileFolder',
   UPLOAD_FILE = 'uploadFile',
   RENAME_FILE_FOLDER = 'renameFileFolder',
+  COPY_FILE_FOLDER = 'copyFileFolder',
```
TomlDev, line 20: "file OR folder" maybe?

libs/src/filesharing/utils/makeUniqueName.ts
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+
+import { DirectoryFileDTO } from '@libs/filesharing/types/directoryFileDTO';
+
+const makeUniqueName = (base: string, ext: string, existingItems: DirectoryFileDTO[]) => {
```
TomlDev, line 15: Please avoid abbr.

libs/src/filesharing/utils/makeUniqueName.ts
```diff
@@ -0,0 +1,39 @@
...
+  const suffixRe = /(.*)\((\d+)\)$/;
```
TomlDev, line 16: Please avoid abbr.

TomlDev (changes requested):
PLease have a look on the comments.
And another thing: Please **always** add descriptions to your PR. It's not trivial what your thoughts were when you implemented your solution.

apps/frontend/src/pages/FileSharing/Dialog/handleFileAction/handleBulkFileOperations.ts
Maddi02: Yes, but I am not quite sure if we should do it. In this case we would use a backend dependency from nestjs move to the frontend.

TomlDev: Thats true, I just found out that we already use it in the shared lib and in the frontend:
grafik

grafik

Could you simply create a new const that we can use in the frontend and replace it in
`apps/frontend/src/store/EduApiStore/useEduApiStore.ts` and
`apps/frontend/src/pages/Surveys/Tables/useSurveysTablesPageStore.ts`
```
export const HttpStatus = {
  CONTINUE: 100,
  SWITCHING_PROTOCOLS: 101,
  PROCESSING: 102,
  EARLY_HINTS: 103,
  OK: 200,
  CREATED: 201,
  ACCEPTED: 202,
  NON_AUTHORITATIVE_INFORMATION: 203,
  NO_CONTENT: 204,
  RESET_CONTENT: 205,
  PARTIAL_CONTENT: 206,
  MULTI_STATUS: 207,
  ALREADY_REPORTED: 208,
  CONTENT_DIFFERENT: 210,
  AMBIGUOUS: 300,
  MOVED_PERMANENTLY: 301,
  FOUND: 302,
  SEE_OTHER: 303,
  NOT_MODIFIED: 304,
  TEMPORARY_REDIRECT: 307,
  PERMANENT_REDIRECT: 308,
  BAD_REQUEST: 400,
  UNAUTHORIZED: 401,
  PAYMENT_REQUIRED: 402,
  FORBIDDEN: 403,
  NOT_FOUND: 404,
  METHOD_NOT_ALLOWED: 405,
  NOT_ACCEPTABLE: 406,
  PROXY_AUTHENTICATION_REQUIRED: 407,
  REQUEST_TIMEOUT: 408,
  CONFLICT: 409,
  GONE: 410,
  LENGTH_REQUIRED: 411,
  PRECONDITION_FAILED: 412,
  PAYLOAD_TOO_LARGE: 413,
  URI_TOO_LONG: 414,
  UNSUPPORTED_MEDIA_TYPE: 415,
  REQUESTED_RANGE_NOT_SATISFIABLE: 416,
  EXPECTATION_FAILED: 417,
  I_AM_A_TEAPOT: 418,
  MISDIRECTED: 421,
  UNPROCESSABLE_ENTITY: 422,
  LOCKED: 423,
  FAILED_DEPENDENCY: 424,
  PRECONDITION_REQUIRED: 428,
  TOO_MANY_REQUESTS: 429,
  UNRECOVERABLE_ERROR: 456,
  INTERNAL_SERVER_ERROR: 500,
  NOT_IMPLEMENTED: 501,
  BAD_GATEWAY: 502,
  SERVICE_UNAVAILABLE: 503,
  GATEWAY_TIMEOUT: 504,
  HTTP_VERSION_NOT_SUPPORTED: 505,
  INSUFFICIENT_STORAGE: 507,
  LOOP_DETECTED: 508,
} as const;
```

Maddi02: Done

apps/api/src/filesharing/consumers/copyFile.consumer.ts
TomlDev: As stated above, `generateUniqueFilename` is not a good name for this function. A typical `generateUniqueFilename` function would not expect a third parameter `items` so this is not a typical `generateUniqueFilename` function. -> There must be a better name.

apps/api/src/filesharing/filesharing.service.ts
TomlDev: `moveOrRenameResources`

apps/frontend/src/pages/FileSharing/Dialog/handleFileAction/handleBulkFileOperations.ts
```diff
@@ -18,35 +18,46 @@ import PathChangeOrCreateDto from '@libs/filesharing/types/pathChangeOrCreatePro
...
-export default handleBulkFileOperations;
+    setResult(success, t('fileOperationSuccessful'), HttpStatus.OK);
+  } catch (rawError: unknown) {
+    const err = rawError as { message?: string; statusCode?: number };
+
+    const message = err.message ?? t('fileOperationFailed');
+    const statusCode = err.statusCode ?? 500;
+
+    setResult(false, message, statusCode);
+  }
```
TomlDev, line 62: This was proposed by the AI, if you want to use the code you should check if it makes sense.
E.g. `fileOperationFailed` is nowhere translated in our project. Also I don't know if `setResult(false, message, statusCode);` makes sense here (or is handled in any way)

apps/frontend/src/hooks/useNotifications.ts
```diff
@@ -144,15 +144,22 @@ const useNotifications = () => {
...
-    const rawFileOperations = handleFilesharingProgress(setFileOperationProgress);
+    const fileOperationHandler = handleFilesharingProgress(setFileOperationProgress);
 
-    const handleFileOperations = (e: MessageEvent<string>) => {
-      void rawFileOperations(e);
+    const listener = (ev: MessageEvent): void => {
+      const typedEvent = ev as MessageEvent<string>;
+      void fileOperationHandler(typedEvent).catch((err) => {
+        console.error('SSE-Error:', err);
+      });
```
TomlDev, line 153: Is the `handleFilesharingProgress` used somewhere else? If not, I think you could omit the wrapper and use it like this:
```
    const listener = (ev: MessageEvent<string>): void => {
      void handleFilesharingProgress(setFileOperationProgress)(ev).catch((err) => {
        console.error('SSE-Error:', err);
      });
    };

**Review by TomlDev — CHANGES_REQUESTED** [review]
Some minor things

File: apps/api/src/filesharing/consumers/copyFile.consumer.ts (hunk 1 above)
Comment (Maddi02) [reply apps/api/src/filesharing/consumers/copyFile.consumer.ts]: What about `getNextAvailableFilename`

File: apps/frontend/src/pages/FileSharing/Dialog/handleFileAction/handleBulkFileOperations.ts (hunk 11 above)
Comment (Maddi02) [reply apps/frontend/src/pages/FileSharing/Dialog/handleFileAction/handleBulkFileOperations.ts:62]: setResult in handleDeleteItems reports the delete operation’s success or failure back to the caller (handleItemAction → setFileOperationResult). I guess we should keep it.

File: apps/frontend/src/hooks/useNotifications.ts (hunk 12 above)
Comment (Maddi02) [reply apps/frontend/src/hooks/useNotifications.ts:153]: Yes, 'handleFilesharingProgress' is used twice because of that. There is now a method called `subscribeFilesharingProgress` which handles all of that in one place.

File: apps/api/src/filesharing/consumers/copyFile.consumer.ts
Comment (TomlDev) [inline apps/api/src/filesharing/consumers/copyFile.consumer.ts:1]: Note related: Wrong text in the copy mdoal
grafik

Comment (TomlDev) [inline apps/api/src/filesharing/consumers/copyFile.consumer.ts:1]: Not related:
Normally I would expect if no folder in the modal is actively selected, it should copy to the current (open) directory.
What about pre-selecting the current (active) directory?
This has the advantage that you dont have to show the current folder in the list for selecting and I think it's very close to normal Windows Explorer behavior

Comment (TomlDev) [inline apps/api/src/filesharing/consumers/copyFile.consumer.ts:1]: In the copy modal the folder where you currently are also appears in the list which is wrong because it allows to navigate inside (which opens again the same folder level):
grafik

Comment (TomlDev) [inline apps/api/src/filesharing/consumers/copyFile.consumer.ts:1]: When copying, it somehow first shows "Fileoperation successful" before showing the progress:
grafik

Comment (TomlDev) [inline apps/api/src/filesharing/consumers/copyFile.consumer.ts:1]: When I copy a whole folder, it does not show a progress toaster, is that on purpose?

https://github.com/user-attachments/assets/62844b9c-6595-4d3f-97ed-71cae3742927

File: apps/api/src/filesharing/consumers/copyFile.consumer.ts (hunk 1 above)
Comment (TomlDev) [reply apps/api/src/filesharing/consumers/copyFile.consumer.ts]: Yes, that sounds good

File: apps/frontend/src/hooks/useNotifications.ts [hunk 13]
```diff
@@ -59,6 +59,29 @@ const useNotifications = () => {
...
+    const asyncHandler = handleFilesharingProgress(setter);
+
+    const listener = (ev: Event) => {
+      void asyncHandler(ev as MessageEvent<string>).catch((err) => {
+        console.error('SSE-Error:', err);
+      });
+    };
+    types.forEach((t) => eventSource?.addEventListener(t, listener, { signal: controller.signal }));
+    return () => types.forEach(() => controller.abort());
+  };
```
TomlDev, line 83: This now seems even more complex than your previous solutions. Maybe you can use the AI for refactoring.
<details>
  <summary>Example, Click to expand</summary>
```
Here’s a much leaner, more reusable pattern that:

1. **Unifies** your two functions into a single `subscribeProgress` helper
2. **Automatically** tears down via one `AbortController.abort()`
3. **Lets you** plug in whatever JSON‐→‐DTO parsing and “clear when done” logic you like

```ts
type ProgressHandler<T> = (data: T) => void | Promise<void>;

/**
 * Subscribes to one or more SSE event types, parses each event as JSON→T,
 * calls your handler, and automatically cleans up when you call the returned
 * unsubscribe() (or if your handler aborts).
 */
function subscribeProgress<T>(
  eventSource: EventSource,
  types: string[],
  handler: ProgressHandler<T>,
): () => void {
  const controller = new AbortController();

  const listener: EventListener = (evt) => {
    const e = evt as MessageEvent<string>;
    let data: T;
    try {
      data = JSON.parse(e.data) as T;
    } catch (err) {
      console.error('Invalid JSON in SSE event', err);
      return;
    }
    // call your handler, catch any rejections so they don’t bubble
    Promise.resolve(handler(data)).catch((err) => {
      console.error('Error in progress handler:', err);
    });
  };

  // attach once per type
  for (const t of types) {
    eventSource.addEventListener(t, listener, { signal: controller.signal });
  }

  // return a single cleanup fn
  return () => {
    controller.abort();
  };
}
```

---

### How you’d use it in your case

```ts
// wherever you need to subscribe:
const unsubscribe = subscribeProgress<FilesharingProgressDto>(
  eventSource!,
  ['upload-progress', 'download-progress'],  // whatever your SSE “type” strings are
  async (data) => {
    setter(data);
    await clearProgressIfComplete(data, setter);
  },
);

// …and when you need to tear down (e.g. in a React useEffect cleanup):
unsubscribe();
```

---

### Bonus: a React hook

If you’re in React-land, you can wrap that in a `useEffect` hook so you never have to write manual cleanup again:

```ts
import { useEffect, useState } from 'react';

function useFileSharingProgress(
  eventSource: EventSource,
  types: string[],
) {
  const [progress, setProgress] = useState<FilesharingProgressDto | null>(null);

  useEffect(() => {
    const unsubscribe = subscribeProgress<FilesharingProgressDto>(
      eventSource,
      types,
      async (data) => {
        setProgress(data);
        await clearProgressIfComplete(data, setProgress);
      },
    );
    return unsubscribe;
  }, [eventSource, types.join(',')]);

  return progress;
}

// Usage in a component:
function MyComponent({ eventSource }: { eventSource: EventSource }) {
  const progress = useFileSharingProgress(eventSource, ['fileshare']);
  // ...
}
```

This gives you a single, well-tested subscription utility, a clean one-line teardown, and (if you use the hook) idiomatic React lifecycle handling.

```
</details>

**Review by TomlDev — CHANGES_REQUESTED** [review]
Please have a look on the comments

File: apps/frontend/src/hooks/subscribeProgress.ts
Comment (michidold) [inline apps/frontend/src/hooks/subscribeProgress.ts:1]: This is not a hook. I would call the useEffect here inside as well and call it useFileOperationProgress.ts

File: apps/frontend/src/hooks/useProgress.ts
Comment (michidold) [inline apps/frontend/src/hooks/useProgress.ts:1]: Seems not be used, was this meant to be called inside useNotifications?

File: apps/frontend/src/pages/FileSharing/Dialog/handleFileAction/handleBulkFileOperations.ts [hunk 14]
```diff
@@ -18,35 +18,41 @@ import PathChangeOrCreateDto from '@libs/filesharing/types/pathChangeOrCreatePro
...
-  });
-};
+async function handleDeleteItems(itemsToDelete: PathChangeOrCreateDto[], endpoint: string): Promise<void> {
+  const cleanPaths = itemsToDelete.map((item) => getPathWithoutWebdav(item.path));
+  const url = buildApiDeletePathUrl(endpoint, DeleteTargetType.FILE_SERVER);
+  await eduApi.delete(url, { data: { paths: cleanPaths } });
+}
 
-const handleBulkFileOperations = async (
+export default async function handleBulkFileOperations(
```
michidold, line 29: Please keep the arrow function.

apps/frontend/src/pages/FileSharing/Dialog/handleFileAction/handleBulkFileOperations.ts
```diff
@@ -18,35 +18,41 @@ import PathChangeOrCreateDto from '@libs/filesharing/types/pathChangeOrCreatePro
...
-    data: {
-      paths: pathsToDelete.map((item) => getPathWithoutWebdav(item.path)),
-    },
-  });
-};
+async function handleDeleteItems(itemsToDelete: PathChangeOrCreateDto[], endpoint: string): Promise<void> {
+  const cleanPaths = itemsToDelete.map((item) => getPathWithoutWebdav(item.path));
+  const url = buildApiDeletePathUrl(endpoint, DeleteTargetType.FILE_SERVER);
+  await eduApi.delete(url, { data: { paths: cleanPaths } });
+}
```
michidold, line 27: Normally we handled api calls inside the store. Why is that here be done?

libs/src/common/constants/httpStatus.ts
michidold, line 1: Think this is not needed since axios has a enum built in:
`import { HttpStatusCode } from 'axios';`

apps/frontend/src/hooks/subscribeProgress.ts
Maddi02: This is not a hook; it makes sense to me. I didn't understand the remaining sentence. Could you take another look? I hope that makes sense.

apps/frontend/src/hooks/useProgress.ts
Maddi02: yes but i removed it

apps/frontend/src/pages/FileSharing/Dialog/handleFileAction/handleBulkFileOperations.ts
Maddi02: To be honest, I don't know why. I'm going to move into the store.

Maddi02: If I understand you correctly, it's about that function that I removed from that file and returned to the store. However, files such as 'handleBulkFileOperations' and 'handleSingleData' are just functions that are called by the store to outsource some code.

apps/frontend/src/hooks/useNotifications.ts
```diff
@@ -129,54 +144,15 @@ const useNotifications = () => {
...
-
-    eventSource.addEventListener(SSE_MESSAGE_TYPE.FILESHARING_DELETE_FILES, handleDelete, { signal });
-
-    return () => {
-      controller.abort();
-    };
-  }, [isFileSharingActive]);
+    if (!isFileSharingActive || !eventSource) return;
+    if (filessaringProgress) setFileOperationProgress(filessaringProgress);
+  }, [isFileSharingActive, eventSource, filessaringProgress]);
```
michidold, line 150: Looks good, but you can handle these two useEffects in the useFileOperationProgress hook at all. So instead of calling two times that hook, you just add two useEffects to useFileOperationProgress.ts and handle everything over there.

michidold (changes requested):
One last finding. Think the useEffects and store logic should be completely handled inside useFileOperationsProgress hook.

apps/frontend/src/pages/FileSharing/Dialog/DialogBodys/dialogBodyConfigurations.ts
```diff
@@ -221,6 +224,35 @@ const dialogBodyConfigurations: Record<string, DialogBodyConfiguration> = {
     },
   },
 
+  copyFileFolder: {
```
TomlDev, line 227: copyFileOrFolder

apps/frontend/src/pages/FileSharing/Dialog/DialogBodys/dialogBodyConfigurations.ts
```diff
@@ -221,6 +224,35 @@ const dialogBodyConfigurations: Record<string, DialogBodyConfiguration> = {
...
+    Component: CopyContentDialogBody,
+    titleKey: 'copyItemDialog.copyFilesToDirectory',
```
TomlDev, line 229: Probably should be `copyFilesOrDirectoriesToDirectory`

apps/api/src/filesharing/consumers/copyFile.consumer.ts
```diff
@@ -0,0 +1,75 @@
...
+    const progressDto: FilesharingProgressDto = {
+      processID: Number(job.id),
+      title: 'filesharing.progressBox.titleCopying',
+      description: 'filesharing.progressBox.fileInfoCopying',
+      statusDescription: 'filesharing.progressBox.processedCopyingInfo',
+      processed,
+      total,
+      percent,
+      currentFilePath: originFilePath,
+      studentName,
```
TomlDev, line 67: I think `studentsName` came from when it was only used in class management but is this still correct?

apps/frontend/src/pages/FileSharing/Dialog/useFileSharingDialogStore.ts
```diff
@@ -109,12 +114,19 @@ const useFileSharingDialogStore = create<FileSharingDialogStore>((set, get) => (
         await handleFileOrCreateFile(action, endpoint, httpMethod, type, data);
         get().setFileOperationResult(true, t('fileOperationSuccessful'), 200);
       } else if (Array.isArray(data)) {
+        const decodedFilenameData = (data as PathChangeOrCreateDto[]).map((d) => ({
```
TomlDev, line 117: `Data` is probably one of the worst variable names because when programming EVERYTHING is data.
`decodedFilenameDtos` instead of `decodedFilenameData`
`bulkDtos` instead of `data`
`dto` instead of `d`

libs/src/filesharing/utils/makeUniqueName.ts
TomlDev, line 1: Filename should be `getNextAvailableFilename`

apps/api/src/filesharing/consumers/copyFile.consumer.ts
TomlDev, line 1: Not related to this file, the folder name is not decoded:
grafik

TomlDev, line 1: Not related to this file: The progress toast is not shown when I copy two files into a folder

https://github.com/user-attachments/assets/5bd6c5da-3a19-4715-84fd-cac9da87c3ca

TomlDev (changes requested):
Toaster is not working.
Other than that only small things to do 👍🏽

apps/api/src/filesharing/consumers/copyFile.consumer.ts
Maddi02: It depends on how long the action takes to complete. Sometimes it's finished even before the progress is shown.

apps/frontend/src/locales/en/translation.json
```diff
@@ -1231,7 +1231,7 @@
     "usersAdded": "{{ count }} added"
   },
   "copyItemDialog": {
-    "copyFilesToDirectory": "Copy files to directory",
+    "copyFilesOrDirectoriesToDirectory": "Copy files to directory",
```
TomlDev, line 1234: Copy files or directories into a directory

apps/frontend/src/locales/de/translation.json
```diff
@@ -1277,7 +1277,7 @@
     "unknownErrorOccurred": "Ein unbekannter Fehler ist aufgetreten"
   },
   "copyItemDialog": {
-    "copyFilesToDirectory": "Dateien in Verzeichnis kopieren",
+    "copyFilesOrDirectoriesToDirectory": "Dateien in Verzeichnis kopieren",
```
TomlDev, line 1280: Dateien oder Verzeichnisse in ein Verzeichnis kopieren

TomlDev (changes requested):
Only some small things, FFTR when solved 👍🏽

michidold (approved):
:+1:

apps/frontend/src/locales/de/translation.json
```diff
@@ -355,12 +355,13 @@
...
-      "fileInfoSharing": "Datei {{filename}} wird an {{studentName}} ausgeteilt",
-      "fileInfoCollecting": "Datei {{filename}} wird von {{studentName}} eingesammelt",
+      "fileInfoSharing": "Datei {{filename}} wird an {{username}} ausgeteilt",
+      "fileInfoCollecting": "Datei {{filename}} wird von {{username}} eingesammelt",
       "fileInfoMoving": "Datei {{filename}} wird verschoben",
       "fileInfoDeleting": "Datei {{filename}} wird gelöscht",
       "fileInfoCopying": "Datei {{filename}} wird kopiert",
-      "errorInfo": "{{failed}} Dateien konnten nicht ausgeteilt werden"
+      "errorInfo": "{{failed}} Dateien konnten nicht ausgeteilt werden",
+      "downloadInfo": "Datei {{filename}} wird heruntergeladen"
```
TomlDev, line 364: These changes are missing in EN translation

apps/frontend/src/hooks/useProgressToast.tsx
TomlDev, line 1: better Filename: `useFileOperationProgressToast`
better hook name: `useFileOperationProgressToast`

TomlDev (approved):
Only 2 small things

PR #799, opened 2025-05-07
665 filemanager rework webdav xml response parsing logic

Maddi02:


libs/src/filesharing/utils/decodeBase64.ts
```diff
@@ -0,0 +1,15 @@
...
+ * This program is free software: you can redistribute it and/or modify it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+
+const fromBase64 = (str: string) => Buffer.from(str, 'base64').toString('utf-8');
```
michidold, line 13: We yet have another base64 decoder which works for frontend only, because of atob/btoa etc. `getBase64String.ts` Maybe you can add this function also to that folder and name it more descriptive like eg decodeBase64Api

libs/src/filesharing/utils/parseWebDAVMultiStatus.ts
```diff
@@ -0,0 +1,44 @@
...
+  textNodeName: '#text',
+  arrayMode: true,
+  parseNodeValue: true,
+  parseAttributeValue: false,
+  trimValues: true,
+};
+
+const parseWebDAVMultiStatus = (xmlData: string) => {
+  const parser = new XMLParser(xmlOptions);
+  const jsonObj = parser.parse(xmlData) as WebdavMultiStatus;
```
michidold, line 29: I would try catch it here, then you don't need to in the parent funcitons mapToDirectories etc

michidold (changes requested):
Found some minor stuff. Please have a look.

libs/src/filesharing/utils/makeUniqueName.ts
michidold, line 1: File could be deleted.

michidold (approved):
Found some small stuff. 

##FFTR

TomlDev (approved):

PR #806, opened 2025-05-09
Fix sidebar in mobile view

TomlDev:
I extracted the classname to give it an explaining variable name.
Also fixed the issue:
grafik

Tested in FF and Chrome, works also for long app names:
grafik


michidold (approved):
:+1:

PR #810, opened 2025-05-12
Fix tooltips in class management

TomlDev:
Before the tooltip was cut off due to the new design where the overflow is hidden. After changing the tooltip position I also changed the background and added a general shadow.
Before:
grafik

After:
grafik

In the GroupListCard I removed the tooltip because it was not useful anymore since long names would still be cut off due to the overflow: hidden. The full name can be seen by opening the dialog.


michidold (approved):
:+1:

PR #811, opened 2025-05-12
796 Fix login page on small heights

TomlDev:
This was the easiest solution I could find.

https://github.com/user-attachments/assets/7c908a9e-9a2e-445f-bcec-1bd69f11c6b3


michidold (approved):
:+1:

PR #816, opened 2025-05-12
Fix tldraw asset handling for non-admins

TomlDev:
It was secured for global-admins only. Moved to it's own controller.


apps/frontend/src/pages/Whiteboard/TLDrawWithSync.tsx
```diff
@@ -83,11 +83,9 @@ const TldrawWithSync = ({ uri, userLanguage, userName }: TldrawWithSyncProps) =>
...
       if (stillUsed) return;
 
-      void eduApi
-        .delete(`/${EDU_API_CONFIG_ENDPOINTS.FILES}/${APPS.WHITEBOARD}/${encodeURIComponent(fileName)}`)
-        .catch((err) => {
-          console.error('Failed to delete file on server', err);
-        });
+      void eduApi.delete(`${assetBasePath}/${encodeURIComponent(fileName)}`).catch((err) => {
+        console.error('Failed to delete file on server', err);
+      });
```
michidold, line 88: This seems to be not handled by tldraw itself, so we could use `handleApiError(error, ()=>{}` here.

apps/frontend/src/pages/Whiteboard/TLDrawWithSync.tsx
```diff
@@ -99,15 +97,15 @@ const TldrawWithSync = ({ uri, userLanguage, userName }: TldrawWithSyncProps) =>
...
+        await eduApi.post<string>(assetPath, form, {
           headers: {
             [HTTP_HEADERS.ContentType]: RequestResponseContentType.MULTIPART_FORM_DATA,
           },
         });
 
-        const url = `${eduApi.defaults.baseURL}${EDU_API_CONFIG_ENDPOINTS.FILES}/${FILE_ENDPOINTS.FILE}/${APPS.WHITEBOARD}/${encodeURIComponent(
-          filename,
-        )}`;
+        const url = `${eduApi.defaults.baseURL}${assetPath}`;
```
michidold, line 108: Its not widely used, I would prefer to just use the constant directly `EDU_API_URL`

michidold (approved):
Only two small remarks. Feel free to resolve them.

apps/frontend/src/pages/Whiteboard/TLDrawWithSync.tsx
TomlDev: Good catch 👍🏽

PR #818, opened 2025-05-13
817 - Check app permissions on endpoint access

michidold:


apps/api/src/auth/access.guard.ts
```diff
@@ -0,0 +1,125 @@
...
+import EVENT_EMITTER_EVENTS from '@libs/appconfig/constants/eventEmitterEvents';
+import getIsAdmin from '@libs/user/utils/getIsAdmin';
+import { APP_ACCESS_KEY, PUBLIC_ROUTE_KEY } from '@libs/auth/constants/appAccessKeys';
+import AppConfigService from '../appconfig/appconfig.service';
+import CustomHttpException from '../common/CustomHttpException';
+import GlobalSettingsService from '../global-settings/global-settings.service';
+
+const accessCache = new Map<string, { value: boolean; expiresAt: number }>();
+const CACHE_TTL_MS = 60_000;
+const MAX_CACHE_SIZE = 10_000;
```
TomlDev: MAX_CACHE_SIZE is defined but never actually used to limit entries. The cleanup only removes expired entries, not excess ones. The cache could grow indefinitely. FFTR

apps/api/src/common/decorators/requireAppAccess.decorator.ts
TomlDev: I think since we can also use it on request level I think we can add it to specific requests of the lmnApi controller that should only be accessible when the user has the class management app:

- [ ] printPasswords
- [ ] startExamMode
- [ ] addManagementGroup
- [ ] removeManagementGroup
- [ ] toggleSchoolClassJoined
- [ ] addUserSession
- [ ] removeUserSession
- [ ] updateUserSession
- [ ] createProject
- [ ] updateProject
- [ ] deleteProject
- [ ] toggleProjectJoined
- [ ] togglePrinterJoined
- [ ] setPassword
- [ ] setFirstPassword

What do you think?

TomlDev (commented):
Very clean and elegant solution 👌🏽

PR #820, opened 2025-05-13
Feature add log level control

michidold:
- Added a env var to control log levels
- Refined loggininterceptor

### DIFF
--- a/apps/api/.env.default
+++ b/apps/api/.env.default
@@ -3,7 +3,7 @@ EDUI_HOST=localhost
 EDUI_PORT=3001
 EDUI_CORS_URL="*"
 
-# Loglevels: off, log, warn, error, debug, verbose, *fatal not used
+# Loglevels: off, log, error, warn, debug, verbose
 EDUI_LOG_LEVEL=
 
 # For file access and onlyoffice
--- a/apps/api/.env.default
+++ b/apps/api/.env.default
@@ -2,6 +2,7 @@
 EDUI_HOST=localhost
 EDUI_PORT=3001
 EDUI_CORS_URL="*"
+EDUI_LOG_LEVEL=debug
 
 # For file access and onlyoffice
 EDUI_WEBDAV_URL=
--- a/apps/api/.env.default
+++ b/apps/api/.env.default
@@ -2,6 +2,8 @@
 EDUI_HOST=localhost
 EDUI_PORT=3001
 EDUI_CORS_URL="*"
+
+# Loglevels: ['log', 'error', 'warn', 'debug', 'verbose']
 EDUI_LOG_LEVEL=debug
 
 # For file access and onlyoffice
--- a/apps/api/.env.default
+++ b/apps/api/.env.default
@@ -3,7 +3,7 @@ EDUI_HOST=localhost
 EDUI_PORT=3001
 EDUI_CORS_URL="*"
 
-# Loglevels: off, log, error, warn, debug, verbose
+# Loglevels: off, error, warn, log, debug, verbose
 EDUI_LOG_LEVEL=
 
 # For file access and onlyoffice
--- a/apps/api/.env.default
+++ b/apps/api/.env.default
@@ -3,7 +3,7 @@ EDUI_HOST=localhost
 EDUI_PORT=3001
 EDUI_CORS_URL="*"
 
-# Loglevels: log, warn, error, debug, verbose, *fatal *not used
+# Loglevels: off, log, warn, error, debug, verbose, *fatal not used
 EDUI_LOG_LEVEL=
 
 # For file access and onlyoffice
--- a/apps/api/.env.default
+++ b/apps/api/.env.default
@@ -3,8 +3,8 @@ EDUI_HOST=localhost
 EDUI_PORT=3001
 EDUI_CORS_URL="*"
 
-# Loglevels: ['log', 'error', 'warn', 'debug', 'verbose']
-EDUI_LOG_LEVEL=debug
+# Loglevels: log, warn, error, debug, verbose, *fatal *not used
+EDUI_LOG_LEVEL=
 
 # For file access and onlyoffice
 EDUI_WEBDAV_URL=
--- a/apps/api/src/logging/getLogLevels.ts
+++ b/apps/api/src/logging/getLogLevels.ts
@@ -13,7 +13,7 @@
 import { LogLevel } from '@nestjs/common';
 import LOG_LEVELS from './log-levels';
 
-const getLogLevels = (envValue: string | undefined): LogLevel[] | false => {
+const getLogLevels = (envValue: string | undefined): LogLevel[] | undefined => {
   const allLevels: LogLevel[] = Object.values(LOG_LEVELS);
   const fallBackLevels = [LOG_LEVELS.ERROR, LOG_LEVELS.WARN, LOG_LEVELS.LOG];
   if (!envValue) {
@@ -21,7 +21,7 @@ const getLogLevels = (envValue: string | undefined): LogLevel[] | false => {
   }
 
   if (envValue === 'off') {
-    return false;
+    return undefined;
   }
 
   const level = envValue.trim().toLowerCase() as LogLevel;
--- a/apps/api/src/logging/getLogLevels.ts
+++ b/apps/api/src/logging/getLogLevels.ts
@@ -1,0 +1,26 @@
+/*
+ * LICENSE
+ *
+ * This program is free software: you can redistribute it and/or modify it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more details.
+ *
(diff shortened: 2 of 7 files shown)


apps/api/src/logging/getLogLevels.ts
```diff
@@ -0,0 +1,26 @@
...
+import { LogLevel } from '@nestjs/common';
+
+const getLogLevels = (envValue: string | undefined): LogLevel[] => {
+  const allLevels: LogLevel[] = ['error', 'warn', 'log', 'debug', 'verbose'];
+  if (!envValue) {
+    return process.env.NODE_ENV === 'production' ? ['error', 'warn'] : ['log', 'error', 'warn', 'debug', 'verbose'];
+  }
+
+  const requested = envValue.split(',').map((l) => l.trim()) as LogLevel[];
+  const valid = requested.filter((l) => allLevels.includes(l));
```
TomlDev, line 22: What do you think about adding a logging output here for all invalid log levels?
```
  const invalid = requested.filter((l) => !allLevels.includes(l as LogLevel));
  if (invalid.length > 0) {
    console.warn(`Ignoring invalid log levels: ${invalid.join(', ')}`, 'getLogLevels');
  }

File: apps/api/src/logging/getLogLevels.ts
Comment (TomlDev) [inline apps/api/src/logging/getLogLevels.ts:1]: What about

> ### Hierarchical (cumulative) levels
> 
> By default the above will only return *exactly* the levels you asked for.  If instead you’d like “debug” to mean “debug **plus** all the more-severe levels” (so that you don’t have to list `error,warn,log,debug` yourself), you can:
> 
> 1. Give each level a numeric rank
> 2. Find the highest‐ranked one requested
> 3. Slice your master list up through that index
> 
> ```ts
> import { LogLevel, Logger } from '@nestjs/common';
> 
> const getLogLevels = (envValue: string | undefined): LogLevel[] => {
>   const allLevels: LogLevel[] = ['error', 'warn', 'log', 'debug', 'verbose'];
>   const rank: Record<LogLevel, number> = {
>     error:   0,
>     warn:    1,
>     log:     2,
>     debug:   3,
>     verbose: 4,
>   };
> 
>   const requested = envValue
>     ? envValue.split(',').map((l) => l.trim()) as string[]
>     : [];
> 
>   const valid = requested.filter((l): l is LogLevel => allLevels.includes(l as LogLevel));
>   const invalid = requested.filter((l) => !allLevels.includes(l as LogLevel));
>   if (invalid.length) {
>     Logger.warn(`Ignoring invalid log levels: ${invalid.join(', ')}`, 'getLogLevels');
>   }
> 
>   // no valid → fallback defaults
>   if (valid.length === 0) {
>     return process.env.NODE_ENV === 'production'
>       ? ['error', 'warn']
>       : allLevels.slice();  // everything
>   }
> 
>   // find “highest” level they asked for
>   const highest = valid
>     .map((lvl) => rank[lvl])
>     .reduce((max, cur) => Math.max(max, cur), 0);
> 
>   // return allLevels up through that rank
>   return allLevels.slice(0, highest + 1);
> };
> ```
> 
> Now:
> 
> * If they set `LOG_LEVEL=debug`, you’ll get `['error','warn','log','debug']`.
> * If they set `LOG_LEVEL=warn,verbose`, you’ll get all the way up to `verbose`.
> * Any typos (e.g. `LOG_LEVEL=foo,bar`) get `Ignoring invalid log levels: foo, bar`…
> 
> ---
> 
> **Does it make sense to have to enumerate every level?**
> 
> Not really – most logging frameworks treat levels *hierarchically*, so that selecting “debug” implicitly gives you all higher-priority messages (error, warn, log) without having to list them by hand.  The slicing approach above is exactly how you can get that behavior in one place.
>

File: apps/api/.env.default [hunk 2]
```diff
@@ -2,6 +2,7 @@
 EDUI_HOST=localhost
 EDUI_PORT=3001
 EDUI_CORS_URL="*"
+EDUI_LOG_LEVEL=debug
```
TomlDev, line 7: EDUI_LOG_LEVELS

apps/api/src/logging/logging.interceptor.ts
TomlDev, line 1: Some considerations: 
- If log level `log` is disabled there is no log that shows that the application has successfully started.
- If `verbose` is disabled I don't see any log messages anymore from requests which was better before. I think we should use the log levels also for "how verbose" these log messages are.
E.g. bring back on loglevel `log`
```
 this.logger.log(
          `${statusCode} ${Date.now() - now}ms ${context.getClass().name}.${context.getHandler().name} ${method} ${url}`,
```
- What is currently the difference between `debug` enabled and disabled?

apps/api/src/logging/logging.interceptor.ts
```diff
@@ -12,36 +12,62 @@
...
+          }),
+          LoggingInterceptor.name,
+        );
+      }),
+      catchError((err: unknown) => {
+        const error = err instanceof Error ? err : new Error(String(err));
 
-        this.logger.log(
-          `${statusCode} ${Date.now() - now}ms ${context.getClass().name}.${context.getHandler().name} ${method} ${url}`,
+        Logger.debug(
```
TomlDev, line 55: Will this also console log when `error` is enabled and `debug` is disabled?

apps/api/.env.default
michidold: Reworked it, its now one level until every other level is logged as well.

apps/api/src/logging/logging.interceptor.ts
michidold: Is reworked now -> debug shows it as before, verbose is more detailed.

apps/api/src/logging/getLogLevels.ts
michidold: I added a fallback if there is a invalid type

apps/api/.env.default
```diff
@@ -3,6 +3,9 @@ EDUI_HOST=localhost
 EDUI_PORT=3001
 EDUI_CORS_URL="*"
 
+# Loglevels: off, log, warn, error, debug, verbose, *fatal not used
+EDUI_LOG_LEVEL=
```
TomlDev, line 7: With loglevel "LOG" I also saw errors and warnings:
grafik

apps/api/.env.default
```diff
@@ -3,6 +3,9 @@ EDUI_HOST=localhost
 EDUI_PORT=3001
 EDUI_CORS_URL="*"
 
+# Loglevels: off, log, warn, error, debug, verbose, *fatal not used
```
TomlDev, line 6: I think the correct order is `log, error, warn`

apps/api/src/logging/logging.interceptor.ts
```diff
@@ -11,37 +11,102 @@
...
+          case LOG_LEVELS.DEBUG:
+            Logger.error(
+              JSON.stringify({
+                method,
+                url,
+                params,
+                query,
+                ip,
+                message: error.message,
+                stack: error.stack,
```
TomlDev, line 96: I think passing the stack only as second parameter to Logger.error is sufficient (otherwise it's there 2 times, one time formatted, one time not:
grafik

Maybe this can be useful if we connect a logging framework in the future but currently I would remove it

apps/api/src/logging/logging.interceptor.ts
```diff
@@ -11,37 +11,102 @@
...
+                params,
+                query,
+                ip,
+                message: error.message,
+              }),
+              undefined,
+              LoggingInterceptor.name,
+            );
+            break;
+          case LOG_LEVELS.DEBUG:
```
TomlDev, line 87: `case LOG_LEVELS.VERBOSE:`
Should be added (otherwise verbose log would be less verbose than debug mode)

TomlDev (changes requested):
Please have a look on the comments, all in all it looks good 👍🏽

apps/api/.env.default
michidold: Thats a feature (bug) in nestjs that exceptions and stuff are getting logged independent of loglevel config. So if something throws its getting logged. This could only be avoided by placing a custom Logger. Think thats not necessary.

TomlDev: With loglevel log you get warnings, I think thats not related to the nestjs implementation:

grafik

With loglevel error you dont see the warnings

TomlDev (changes requested):
Building api fails:
```
> nx run api:build

> webpack-cli build --node-env=production

chunk (runtime: main) main.js (main) 683 KiB [entry] [rendered]

ERROR in ../../libs/src/common/types/http-methods.ts:56:3
TS1117: An object literal cannot have multiple properties with the same name.
    54 |   XApiKey: 'x-api-key',
    55 |   CONNECTION_UID: 'Connection-Uid',
  > 56 |   ContentLength: 'Content-Length',
       |   ^^^^^^^^^^^^^
    57 |   XForwaredFor: 'x-forwarded-for',
    58 |   UserAgent: 'User-Agent',
    59 | } as const;

webpack compiled with 1 error (42f4ff76d96aff89)
Warning: command "webpack-cli build --node-env=production" exited with non-zero status code
———————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————

 NX   Ran target build for project api (10s)

   ✖  1/1 failed
   ✔  0/1 succeeded [0 read from cache]

Build failed, waiting for changes to restart...

TomlDev (approved):

PR #821, opened 2025-05-13
698 Settings Page: Make tabs header sticky

TomlDev:
After: 

https://github.com/user-attachments/assets/706fc02d-cc20-4d47-861d-d356aef377ea


michidold (changes requested):
Looks good, can you fix it for mobile as well?

michidold (approved):
:+1:

PR #825, opened 2025-05-14
824 File Manager: Fix docked preview is above dialog

TomlDev:
Currently this fix would be the easiest. If the window is docked, it's always behind the dialogs.
If the modal is undocked it will always be in front. You will first have to minimize it since the dialogs don't work like the resizable windows.

After:

https://github.com/user-attachments/assets/62c1aeb5-9bcf-4065-beaf-adb9ed23c684


michidold (approved):
:+1:

PR #836, opened 2025-05-14
Fix exception when license check fails

michidold:
- Catched a error due to license validity check
- Used a logger, since a thrown exception in the onModuleInit prevents the application from starting


Maddi02 (approved):
👍

PR #841, opened 2025-05-16
704 Add health check for docker api

michidold:


apps/api/src/health/health.controller.ts
TomlDev, line 1: 2 endpoints, 1 dedicated for docker with less checks

TomlDev (approved):
👌🏽

PR #850, opened 2025-05-20
672 Add default landing page

TomlDev:
- Remove unnecessary imports of MigrationService (it has only 1 static method, so it has not to be injected)
- Add better logging in DockerService
- Add `generalSettings` with `defaultLandingPage` options to general settings
- Add migration to add the new fields
- Add a new `<DefaultLandingPageAfterLogin>` component that is the first page after the login
- Extract the `<AppDropdownSelectFormField>` component for reusability
- Add the possibility to configure the default landing page in the admin settings

### DIFF
--- a/apps/frontend/src/components/structure/DefaultLandingPageAfterLogin.tsx
+++ b/apps/frontend/src/components/structure/DefaultLandingPageAfterLogin.tsx
@@ -1,0 +1,43 @@
+/*
+ * LICENSE
+ *
+ * This program is free software: you can redistribute it and/or modify it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+
+import React, { useEffect } from 'react';
+import useGlobalSettingsApiStore from '@/pages/Settings/GlobalSettings/useGlobalSettingsApiStore';
+import { useNavigate } from 'react-router-dom';
+import CircleLoader from '@/components/ui/Loading/CircleLoader';
+
+const DefaultLandingPageAfterLogin = () => {
+  const { globalSettings } = useGlobalSettingsApiStore();
+  const navigate = useNavigate();
+  const { isCustomLandingPageEnabled, appName } = globalSettings.general.defaultLandingPage;
+
+  useEffect(() => {
+    if (isCustomLandingPageEnabled === undefined) return;
+
+    if (isCustomLandingPageEnabled) {
+      navigate(`/${appName}`, { replace: true });
+    } else {
+      navigate('/', { replace: true });
+    }
+  }, [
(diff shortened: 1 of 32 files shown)


apps/frontend/src/components/ui/DropdownSelect/AppDropdownSelectFormField.tsx
```diff
@@ -0,0 +1,68 @@
...
+      defaultValue={initialValue}
+      render={({ field }) => (
+        <FormItem>
+          <p className="font-bold">{t('common.application')}</p>
+          <FormControl>
+            <DropdownSelect
+              options={appNameOptions}
+              selectedVal={field.value}
+              handleChange={field.onChange}
+              variant="dialog"
```
michidold, line 59: The variant should be a property. On global-settings page variant = 'default'

apps/frontend/src/pages/Settings/GlobalSettings/GlobalSettings.tsx
```diff
@@ -12,95 +12,138 @@
...
+          <form onSubmit={handleSubmit(onSubmit)}>
+            <AccordionItem value="general">
+              <AccordionTrigger className="flex text-h4">
+                <h4>{t('settings.globalSettings.general')}</h4>
+              </AccordionTrigger>
+              <AccordionContent
+                className="space-y-2 px-1 text-p"
+                style={{ overflow: 'visible' }}
+              >
+                <h4>{t('settings.globalSettings.defaultLandingPageTitle')}</h4>
```
michidold, line 109: Would use a `<p className="text-xl font-bold">` to separate from the accordion trigger text

michidold (changes requested):
Found some small stuff. Please have a look.

apps/frontend/src/components/structure/DefaultLandingPageAfterLogin.tsx
```diff
@@ -0,0 +1,43 @@
...
+  ]);
+
+  if (isCustomLandingPageEnabled === undefined) {
+    return <CircleLoader className="absolute left-1/2 top-1/2 -translate-x-1/2 -translate-y-1/2" />;
+  }
+
+  return null;
+};
+
+export default DefaultLandingPageAfterLogin;
```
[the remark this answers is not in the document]
TomlDev: The undefined state is used as loading state. It shows that the data was not fetched from the backend, so this would lead to a different behavior.
When do you expect `general` to be empty? We set it with the migration in the backend and we load the defaultValues in the store so it is never empty.

apps/frontend/src/components/ui/DropdownSelect/DropdownSelect.tsx
```diff
@@ -103,7 +103,7 @@ const DropdownSelect: React.FC<DropdownProps> = ({
       </div>
       <div className={cn(styles.arrow, { [styles.open]: isOpen, [styles.up]: openToTop })} />
       <div
-        className={cn('scrollbar-thin', styles.options, {
+        className={cn('shadow-xl scrollbar-thin', styles.options, {
```
TomlDev, line 106: Added this shadow to fix this:
grafik

apps/frontend/src/components/structure/DefaultLandingPageAfterLogin.tsx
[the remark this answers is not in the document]
michidold: Persistence in the store was intended for slow queries. Since these settings don't change that often, you could reactivate it.

The interception of the empty object was only intended as a best practice, but you're right it should not happen.

michidold (approved):
:+1:

PR #855, opened 2025-05-20
844 Login: When no app is configured for your access group, the login fails

TomlDev:
Test by switching to a new DB and try to connect as non-global-admin user.


michidold (approved):
:+1:

PR #860, opened 2025-05-20
Router: Fix unknown routes

### LINKED ISSUE #830: Home Route: Replace magic strings
Image

In many places we use `'/'`. This should be replaced where it is used as home path.

TomlDev:
Actual fix is only in `apps/frontend/src/router/routes/AuthRoutes.tsx`

Also fixes #830


michidold (approved):
:+1:

Issue #770, opened 2025-05-02
Survey Templates - add a delete dialog to manage the created templates


maierdominik:
Currently we have to Clear the Template folder by hand.
To simplify the clean up for the global-admin we should create a Dialog were he can select those Templates that should be destroyed

PR #868, opened 2025-05-23
770 - Survey Templates - add a delete dialog to manage the created templates

maierdominik:
Issue: https://github.com/edulution-io/edulution-ui/issues/770

770:
* survey templates add delete button to template dialog

20250522_SurveyTemplates_AddDeleteButton

### DIFF
--- a/apps/frontend/src/pages/Surveys/Editor/dialog/TemplateItem.tsx
+++ b/apps/frontend/src/pages/Surveys/Editor/dialog/TemplateItem.tsx
@@ -17,6 +17,7 @@ import { SurveyCreator } from 'survey-creator-react';
 import cn from '@libs/common/utils/className';
 import SurveyDto from '@libs/survey/types/api/survey.dto';
 import SurveyTemplateDto from '@libs/survey/types/api/template.dto';
+import useLdapGroups from '@/hooks/useLdapGroups';
 import useTemplateMenuStore from '@/pages/Surveys/Editor/dialog/useTemplateMenuStore';
 import { Button } from '@/components/shared/Button';
 import { Textarea } from '@/components/ui/Textarea';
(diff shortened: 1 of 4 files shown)


apps/frontend/src/pages/Surveys/Editor/dialog/TemplateItem.tsx
TomlDev, line 1: Am I doing something wrong? It does not seem to delete the template:
[Bildschirmaufzeichnung vom 2025-05-27 10-34-00.webm](https://github.com/user-attachments/assets/fd79fc2a-8263-4308-9492-e6128c4f1b08)

TomlDev (changes requested):
Seems not to work

apps/frontend/src/pages/Surveys/Editor/dialog/TemplateItem.tsx
maierdominik: It works for me. Without any errors and 
the Log message in the backend also confirms the deletion of the file.

The only thing i could think of is that we are using system specifique paths.
I checked if we are building paths using `string.join('/')` instead of `path.join( ...<string[]> )`.

image


Could you test it again and if necessary it throws an error then please provide the log from the backend

apps/frontend/src/pages/Surveys/Editor/dialog/useTemplateMenuStore.tsx
```diff
@@ -27,6 +29,8 @@ interface TemplateMenuStore {
   uploadTemplate: (template: SurveyTemplateDto) => Promise<void>;
   isSubmitting: boolean;
 
+  deleteTemplate: (templateName: string) => Promise<void>;
```
TomlDev, line 32: `templateFileName: string`

apps/frontend/src/pages/Surveys/Editor/dialog/useTemplateMenuStore.tsx
```diff
@@ -100,6 +104,26 @@ const useTemplateMenuStore = create<TemplateMenuStore>((set) => ({
       set({ isSubmitting: false });
     }
   },
+
+  deleteTemplate: async (templateName: string): Promise<void> => {
```
TomlDev, line 108: `templateFileName: string`

TomlDev (approved):

michidold (changes requested):
Found some code improvements, please have a look.

PR #874, opened 2025-05-27
Implement SafeOnModuleInit decorator to enhance error handling in onM…

Maddi02:
…oduleInit method


TomlDev:
As discussed we should take care that the app only starts if there is no error in module initialization.
Maybe we can use it later for modules that are allowed to throw.

Issue #556, opened 2025-03-03
Survey Result - Fix the styling for the result table and visualization components


maierdominik:
This was previously a part of the following Issue: https://github.com/edulution-io/edulution-ui/issues/495.
Got seperated after some changes are required to get to those results

Survey Result

![Image](https://github.com/user-attachments/assets/afda97fa-ff0b-47af-a62d-cca43b8dedab)

TomlDev:
![Image](https://github.com/user-attachments/assets/1f188f20-9826-4890-a70b-8fe54f209c29)

PR #895, opened 2025-06-03
556 - Survey Result - Fix the styling for the result chart component (NEW)

maierdominik:
Issue: https://github.com/edulution-io/edulution-ui/issues/556

556:
* fix styling for survey result visualization

BEFORE:
image

AFTER:
image


TomlDev (approved):

apps/frontend/src/pages/Surveys/Tables/dialogs/resultVisualizationDialog.css
```diff
@@ -0,0 +1,78 @@
...
+.main-svg,
+.svg-container,
+.user-select-none {
+  background: transparent !important;
+  background-color: transparent !important;
+}
+text {
+  color: var(--primary-foreground) !important;
+  fill: var(--primary-foreground) !important;
+}
```
michidold, line 41: Here global classes are overwritten. This should not be done, since its very hard to debug.

apps/frontend/src/pages/Surveys/Tables/dialogs/resultVisualizationDialog.css
michidold, line 1: There are a lot of `!important` statements. Be careful since the override every global styling.

michidold (changes requested):
Found one issue. Please have a look.

apps/frontend/src/pages/Surveys/Tables/dialogs/resultVisualizationDialog.css
maierdominik: I removed all and only added those that seem necessary (which nearly all of them are^^)

PR #897, opened 2025-06-03
892 - Fix users search limits to 100 

michidold:
- Add a max query param to fetch all users from Keycloak
- Cache prepared user objects in Redis to save space
- Run `npm audit fix`
- Add line `"isolatedModules": true` to tsconfig.spec.json, which was recommended by jest


apps/api/src/users/users.service.ts
```diff
@@ -94,33 +103,30 @@ class UsersService {
     );
 
     await Promise.all(
-      Object.entries(usersBySchool).map(([school, userList]) =>
-        this.cacheManager.set(ALL_USERS_CACHE_KEY + school, userList, DEFAULT_CACHE_TTL_MS),
-      ),
+      Object.entries(usersBySchool).map(async ([school, userList]) => {
+        const cachedList: CachedUser[] = userList.map(mapToCachedUser);
```
TomlDev, line 107: Wouldn't it make sense to map over the users when they are fetched? I think we only have to do it once then.

TomlDev (changes requested):
Looks good, only small things 👍🏽

PR #918, opened 2025-06-06
916 - Add reset mfa functionality

michidold:


apps/frontend/src/pages/Settings/components/UserAdministration.tsx
TomlDev, line 1: I am unsure if it wouldn't be better to encapsulate the resetMfaForm from the general `UserAdministration`.
The values cannot be reused (changes from another component on `selectedUsers` will always change the value for the MFA part too).
Also I would in general move this MFA part to its own component, otherwise if we add more functionality to the user administration this file would grow and grow. Probably a general `form` is not necessary here because we don't send it "as one" to our backend.

TomlDev (changes requested):
Looks good, only some considerations 👍🏽

apps/frontend/src/pages/Settings/components/SettingsOverviewPage.tsx
TomlDev, line 1: It looks like this for me (FF & Chrome)
image

michidold: Yeah, worked at reload due to wrong useMemo usage. Should be fixed now.

TomlDev (approved):

PR #928, opened 2025-06-11
Bulletin Board: Fix word-break, date and color

TomlDev:
- Fix word break
- Enable color and background for text
- Fix problem when you edit an existing bulletin with a start or end date set (you could not save it again)

### DIFF
--- a/apps/frontend/src/locales/de/translation.json
+++ b/apps/frontend/src/locales/de/translation.json
@@ -1235,7 +1235,7 @@
     "cannotUploadOversized": "Es können keine Dateien größer als 50MB hochgeladen werden."
   },
   "warnings": {
-    "colorMayConflictWithTheme": "Manuelle Einstellung der Farbe kann zu Problemen mit anderen Themes (hell/dunkel) führen."
+    "colorMayConflictWithTheme": "Manuelle Einstellung der Farbe kann zukünftig zu Problemen mit anderen Themes (hell/dunkel) führen. Es wird empfohlen, die Farbe der Schrift immer gleichzeitig mit der Hintergrundfarbe zu setzen."
   },
   "linuxmuster": {
     "title": "LINUXMUSTER",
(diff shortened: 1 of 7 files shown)


apps/frontend/src/locales/de/translation.json
```diff
@@ -1234,6 +1234,9 @@
     "oversizedFileDetected": "Zu große Datei entdeckt!",
     "cannotUploadOversized": "Es können keine Dateien größer als 50MB hochgeladen werden."
   },
+  "warnings": {
+    "colorMayConflictWithTheme": "Manuelle Einstellung der Farbe kann zu Problemen mit anderen Themes (hell/dunkel) führen."
```
Maddi02, line 1238: I'm not sure if this makes sense. Currently, we only support dark mode, so this could be misleading. In addition, what will happen if we have more themes that differ from light or dark? We could check how many themes we currently have and, depending on that number, show the current warning to prevent misleading.

apps/frontend/src/locales/en/translation.json
```diff
@@ -1230,6 +1230,9 @@
     "oversizedFileDetected": "Oversized file detected!",
     "cannotUploadOversized": "Cannot upload oversized files (more than 50 MB)."
   },
+  "warnings": {
+    "colorMayConflictWithTheme": "Manual setting of the color can lead to problems with other themes (light/dark)."
```
Maddi02, line 1234: Same as above

Maddi02 (changes requested):
Please take a look at my comment on the translation text. Perhaps we could discuss it.

apps/frontend/src/locales/de/translation.json
TomlDev: What about 
> "Manuelle Einstellung der Farbe kann zukünftig zu Problemen mit anderen Themes (hell/dunkel) führen. Es wird empfohlen, die Farbe der Schrift immer gleichzeitig mit der Hintergrundfarbe zu setzen"

PR #930, opened 2025-06-12
Add app titles to sidebar

TomlDev:
- Add `DynamicEllipsis` which is used to truncate text from the middle ("Schwarzes Brett" -> "Schwa... Brett")
  - It creates a hidden "measure" element to compute the available space for the text
  - Then it truncates until it finds a display text that fits
- Add the app title under the app icon in the desktop sidebar, truncate if necessary
- Add hover effect that scales the icons
- Increase the desktop sidebar width
- Change translations (app names) according to requirements of MC


apps/frontend/src/components/shared/DynamicEllipsis.tsx
```diff
@@ -0,0 +1,91 @@
...
+    const removedCount = text.length - totalVisible;
+    if (removedCount <= 2) {
+      setTruncatedTextToDisplay(text);
+      return;
+    }
+
+    const head = Math.ceil(totalVisible / 2);
+    const tail = totalVisible - head;
+    setTruncatedTextToDisplay(`${text.slice(0, head)  }…${  text.slice(-tail)}`);
+  }, [text, className]);
```
Maddi02, line 73: Since this is only triggered again if the text or class name changes, we would recalculate in the case of resizing, for example. In the sidebar, I guess this should be unnecessary. However, when it comes to reusing this item, this case is missing. What do you think about adding an event listener for resizing and triggering it again afterwards?

Maddi02 (approved):
It looks good. However, I am curious about the resizing point mentioned in the comment. It might be worth talking about => FFTR

PR #933, opened 2025-06-12
932 Sidebar: Items should have a fixed height

TomlDev:
Before:
grafik

After:

https://github.com/user-attachments/assets/ed83caa1-8ed3-49c2-b358-70f89f5c051b


Maddi02 (approved):
👍

PR #946, opened 2025-06-18
 Add the option to add an icon to the adaptive dialog title

Maddi02:
- Added the possibility to add an icon to the adaptive dialog. Added a prop to the AdaptiveDialog in order to show an icon on the left side of the sheet or dialog title.
- In addition, a prop to disable the close action for the dialog was added. This is needed if we use this dialog on a publicly accessible page where the only component with a render is the adaptive dialog.

image
image


apps/frontend/src/components/ui/AdaptiveDialog.tsx
```diff
@@ -41,6 +42,8 @@ interface AdaptiveDialogProps {
   variant?: 'primary' | 'secondary' | 'tertiary';
   mobileContentClassName?: string;
   desktopContentClassName?: string;
+  TitleIcon?: React.ComponentType<IconBaseProps>;
```
TomlDev, line 45: titleIcon

apps/frontend/src/components/ui/AdaptiveDialog.tsx
```diff
@@ -41,6 +42,8 @@ interface AdaptiveDialogProps {
...
+  persistent?: boolean;
```
TomlDev, line 46: What about using the presence of `handleOpenChange` for this? If we make it optional we could use it the same way. 
If it's there we handle the closing, if not, we treat the dialog as persistent. This way we avoid adding more and more props.

Also this behavior is currently not used in your PR?

TomlDev (changes requested):
Please have a look

apps/frontend/src/components/ui/AdaptiveDialog.tsx
```diff
@@ -53,9 +55,27 @@ const AdaptiveDialog: FC<AdaptiveDialogProps> = ({
   variant = 'primary',
   mobileContentClassName,
   desktopContentClassName,
+  titleIcon: TitleIcon,
 }) => {
   const { isMobileView } = useMedia();
 
+  const iconContextValue = useMemo(() => ({ className: 'h-8 w-8 m-5' }), []);
+
+  const isHandleChangesEmpty = (fn?: (() => void) | undefined) => !fn || fn.toString() === '() => {}';
```
TomlDev, line 64: Better make the prop optional
```
interface AdaptiveDialogProps {
  isOpen: boolean;
  handleOpenChange?: () => void;

**Review by TomlDev — CHANGES_REQUESTED** [review]

**Review by TomlDev — CHANGES_REQUESTED** [review]
Ass discussed

**Review by TomlDev — APPROVED** [review]

**Review by michidold — APPROVED** [review]
:+1:

File: apps/frontend/src/components/ui/Sheet.tsx [hunk 4]
```diff
@@ -86,16 +87,18 @@ const SheetContent = React.forwardRef<React.ElementRef<typeof SheetPrimitive.Con
...
+            className={cn(
+              { 'text-card-foreground': variant === 'primary' },
+              { 'text-background': variant === 'secondary' || variant === 'tertiary' },
+              'absolute right-4 top-4 rounded-sm opacity-70 ring-offset-background transition-opacity data-[state=open]:bg-secondary hover:opacity-100 focus:outline-none focus:ring-2 focus:ring-ring focus:ring-offset-2 disabled:pointer-events-none',
+            )}
+          >
+            <Cross2Icon className="h-4 w-4" />
+            <span className="sr-only">{i18n.t('dialog.close')}</span>
+          </SheetPrimitive.Close>
+        )}
```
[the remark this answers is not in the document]
Maddi02: As far as I can tell, it's unfortunately not.

PR #947, opened 2025-06-20
Add launcher with CTRL+K

TomlDev:
- Add an app launcher that can be opened by clicking on the edulution logo or with CTRL/CMD + K
- Display apps in a grid in the launcher
- Make the apps filterable and launch the first app in the list with enter
- Also for mobile view
- Add new hook to determine the CTRL (windows, linux) or CMD key (mac) `useModKeyLabel`
- Add new hook to make the sidebar entries available `useSidebarItems`

### DIFF
--- a/apps/frontend/src/components/ui/Launcher/Launcher.tsx
+++ b/apps/frontend/src/components/ui/Launcher/Launcher.tsx
@@ -1,0 +1,51 @@
+/*
+ * LICENSE
+ *
+ * This program is free software: you can redistribute it and/or modify it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+
+import React, { useCallback, useEffect } from 'react';
+import AdaptiveDialog from '@/components/ui/AdaptiveDialog';
+import useLauncherStore from '@/components/ui/Launcher/useLauncherStore';
+import LauncherAppGrid from '@/components/ui/Launcher/LauncherAppGrid';
+import useModKeyLabel from '@/hooks/useModKeyLabel';
+
+const Launcher = () => {
+  const { isLauncherOpen, toggleLauncher } = useLauncherStore();
+  const modKeyLabel = useModKeyLabel();
+
+  const handleKeyDown = useCallback(
+    (e: KeyboardEvent) => {
+      const isMod = e.ctrlKey || e.metaKey;
+      if (isMod && e.key.toLowerCase() === 'k') {
+        e.preventDefault();
+        toggleLauncher();
+      }
+    },
+    [toggleLauncher],
+  );
+
+  useEffect(() => {
+    window.addEventListener('keydown', handleKeyDown);
+    return () => window.removeEventListener('keydown', handleKeyDown);
+  }, [handleKeyDown]);
+
+  return (
+    <AdaptiveDialog
+      isOpen={isLauncherOpen}
+      handleOpenChange={toggleLauncher}
+      title=""
+      desktopContentClassName="max-h-[95%] max-w-[90%] z-[999]"
+      mobileContentClassName="h-full z-[999] flex flex-col pb-0"
+      body={<LauncherAppGrid modKeyLabel={modKeyLabel} />}
+    />
+  );
+};
+
+export default Launcher;
--- a/apps/frontend/src/components/ui/Launcher/LauncherAppGrid.tsx
+++ b/apps/frontend/src/components/ui/Launcher/LauncherAppGrid.tsx
@@ -22,6 +22,7 @@ import Input from '@/components/shared/Input';
 import isSubsequence from '@libs/common/utils/string/isSubsequence';
 import useMedia from '@/hooks/useMedia';
 import cn from '@libs/common/utils/className';
+import NotificationCounter from '@/components/ui/Sidebar/SidebarMenuItems/NotificationCounter';
 import SEARCH_INPUT_LABEL from '@libs/ui/constants/launcherSearchInputLabel';
 
 const LauncherAppGrid = ({ modKeyLabel }: { modKeyLabel: string }) => {
@@ -87,7 +88,7 @@ const LauncherAppGrid = ({ modKeyLabel }: { modKeyLabel: string }) => {
 
       <div
         className="mx-auto grid max-h-[full] w-full grid-cols-[repeat(auto-fit,minmax(8rem,auto))] justify-center
-        gap-x-3 gap-y-2 overflow-auto overflow-y-auto pb-10 scrollbar-thin md:max-h-full
+        gap-x-3 gap-y-2 overflow-auto pb-10 scrollbar-thin md:max-h-full
         md:w-[95%] md:grid-cols-[repeat(auto-fit,minmax(12rem,auto))] md:gap-x-6 md:gap-y-5 md:pb-4"
       >
         {filteredApps.length ? (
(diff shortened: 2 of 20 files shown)


apps/frontend/src/components/ui/Launcher/LauncherAppGrid.tsx
```diff
@@ -0,0 +1,135 @@
...
+  }, [sidebarItems, language, search]);
+
+  const onClose = useCallback(() => {
+    toggleMobileSidebar();
+    toggleLauncher();
+  }, [toggleMobileSidebar, toggleLauncher]);
+
+  const searchInputLabel = 'launcher.searchApps';
+
+  useEffect(() => {
```
Maddi02, line 55: what about using in here `useCallback`

apps/frontend/src/components/ui/Launcher/LauncherAppGrid.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+      <div
+        className="mx-auto grid max-h-[full] w-full grid-cols-[repeat(auto-fit,minmax(8rem,auto))] justify-center
+        gap-x-3 gap-y-2 overflow-auto overflow-y-auto pb-10 scrollbar-thin md:max-h-full
+        md:w-[95%] md:grid-cols-[repeat(auto-fit,minmax(12rem,auto))] md:gap-x-6 md:gap-y-5 md:pb-4"
+      >
+        {filteredApps.length ? (
+          filteredApps.map((app, index) => (
+            <NavLink
+              key={app.title}
```
Maddi02, line 91: Perhaps we should use the `app.link` here to prevent duplicate keys if the app name is the same.

apps/frontend/src/components/ui/Launcher/LauncherAppGrid.tsx
```diff
@@ -0,0 +1,135 @@
...
+  useEffect(() => {
+    const onKeyDown = (event: KeyboardEvent) => {
+      if (event.key !== 'Enter') return;
+
+      const active = document.activeElement;
+
+      if (active instanceof HTMLInputElement && active.getAttribute('aria-label') === searchInputLabel) {
+        event.preventDefault();
+        onClose();
+        navigate(filteredApps[0].link);
```
Maddi02, line 64: To prevent errors, I guess we should either use the 'at' function or use a '?'.

apps/frontend/src/components/ui/Sidebar/SidebarMenuItems/LauncherButton.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { SIDEBAR_ICON_WIDTH } from '@libs/ui/constants';
+import { useTranslation } from 'react-i18next';
+import useLauncherStore from '@/components/ui/Launcher/useLauncherStore';
+
+const LauncherButton: React.FC = () => {
+  const { t } = useTranslation();
+  const { toggleLauncher } = useLauncherStore();
+
+  return (
+    <button
```
Maddi02, line 24: Is there a reason for using this button instead of our Button component?

apps/frontend/src/hooks/useModKeyLabel.ts
```diff
@@ -0,0 +1,32 @@
...
+import { useEffect, useState } from 'react';
+import { useTranslation } from 'react-i18next';
+
+export default function useModKeyLabel() {
+  const { t } = useTranslation();
+  const [label, setLabel] = useState<string>('STRG');
+
+  useEffect(() => {
+    const platform = navigator.platform || navigator.userAgent || '';
+    const isMac = /Mac/i.test(platform);
```
Maddi02, line 22: What about creating a hook for this, such as `useOperatingSystem`?

apps/frontend/src/locales/de/translation.json
```diff
@@ -759,7 +759,15 @@
     "yourPassword": "Dein Passwort",
     "addButton": "Hinzufügen"
   },
-
+  "ctrlKey": "STRG",
+  "enterKey": "Enter",
+  "launcher": {
+    "title": "App Launcher",
+    "searchApps": "Tippen um Apps zu filtern...",
```
Maddi02, line 766: What about `Tippe um nach Apps zu suchen`

Maddi02 (changes requested):
It works well 👍. I found some minor issues. Feel free to reach out to me if you have any other comments about the things I've mentioned.

apps/frontend/src/components/ui/Launcher/Launcher.tsx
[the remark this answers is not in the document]
TomlDev: You mean that it overflows the sidebar? Should be fine I think

apps/frontend/src/components/ui/Launcher/LauncherAppGrid.tsx
```diff
@@ -0,0 +1,135 @@
...
+      return isSubsequence(searchString, name);
+    });
+  }, [sidebarItems, language, search]);
+
+  const onClose = useCallback(() => {
+    toggleMobileSidebar();
+    toggleLauncher();
+  }, [toggleMobileSidebar, toggleLauncher]);
+
+  const searchInputLabel = 'launcher.searchApps';
```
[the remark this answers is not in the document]
TomlDev: Moved to shared lib

apps/frontend/src/components/ui/Launcher/LauncherAppGrid.tsx
TomlDev: Title and link are both created from the app.name, and they are both unique. Changed it nevertheless to link 😸

apps/frontend/src/components/ui/Sidebar/SidebarMenuItems/LauncherButton.tsx
TomlDev: Yes, we use `<Button>` when we need our variants which is not the case here

apps/frontend/src/hooks/useModKeyLabel.ts
TomlDev: Not necessary at this point, YAGNI

apps/frontend/src/locales/de/translation.json
TomlDev: Hmm I am unsure, because this is more a filtering. The results (apps) are already displayed and then only filtered.

libs/src/common/utils/string/isSubsequence.ts
```diff
@@ -0,0 +1,27 @@
...
+ * This program is free software: you can redistribute it and/or modify it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+
+const isSubsequence = (pattern: string, text: string): boolean => {
```
michidold, line 13: Works well so far. Maybe we could use a fuzzy search lib like fuse.js, which also finds stuff with typos.

##FFTR

michidold (approved):
Found only minor stuff. FFTM after resolve

apps/frontend/src/components/ui/Launcher/Launcher.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+import React, { useCallback, useEffect } from 'react';
+import AdaptiveDialog from '@/components/ui/AdaptiveDialog';
+import useLauncherStore from '@/components/ui/Launcher/useLauncherStore';
+import LauncherAppGrid from '@/components/ui/Launcher/LauncherAppGrid';
+import useModKeyLabel from '@/hooks/useModKeyLabel';
+
+const Launcher = () => {
+  const { isLauncherOpen, toggleLauncher } = useLauncherStore();
+  const modKeyLabel = useModKeyLabel();
```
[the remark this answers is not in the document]
TomlDev: Had this before, but this will reevaluate each time you open the launcher. There is a little delay, so for a moment you see the default value before it switches to your correct one. Since this does not change as long as the application is open, it makes sense to compute it only once here and then pass it every time to the grid.

libs/src/common/utils/string/isSubsequence.ts
TomlDev: https://github.com/edulution-io/edulution-ui/issues/966

PR #992, opened 2025-07-01
Refactor: Renamed getUser decorator file

TomlDev:
The file had the wrong name, the decorator is called `getCurrentUser`


michidold (approved):
:+1:

PR #996, opened 2025-07-01
Login Page: Extract auth error hook, add more translations

TomlDev:
Creates a custom hook so the file does not grow.
Adds a new translation for a backend error response.


apps/frontend/src/pages/LoginPage/useAuthErrorHandler.ts
```diff
@@ -0,0 +1,52 @@
...
+      form.setError('password' as Path<TFormValues>, { message: t('auth.errors.TokenExpired') });
+      return;
+    }
+
+    form.setError('password' as Path<TFormValues>, { message: t(authError.message) });
+
+    if (showQrCode) {
+      toast.error(t(authError.message));
+    }
+  }, [authError, form, showQrCode]);
```
michidold, line 49: With showQrCode in the effect dependencies the toast also gets triggered if there is an old error in the authError state:

[output.webm](https://github.com/user-attachments/assets/a90d2baf-531f-45bf-9d49-231ccd74463e)

I would reset the form when click on showQrCode, so there are no old messages.

michidold (changes requested):
Found one issue. Please have a look.

michidold (approved):
:+1:

PR #1035, opened 2025-07-07
Added two missing roles

michidold:
- Added two missing roles


TomlDev (approved):
The roles seem to be added successfully, I cannot say something for the rest. 👍🏽

PR #1044, opened 2025-07-08
Fix blocking iframe

TomlDev:
I was able to identify a bug that was caused when the `<FramePlaceholder>` component did not unset the active frame when unmounted.
This could happen when you where in a framed app and opened another app in a new tab.
How to reproduce: -> Open "Mail" -> Middle mouse click "Conferences" -> "Mail" is still open and does not close on navigation

Solution was to move everything out of this component and place the logic in a component that does not unmount.


michidold (changes requested):
If I test it local or on dev2 server I get no iframe at all to show up.

PR #1045, opened 2025-07-09
Reload UI if the version changes

TomlDev:
Add toaster to refresh the page:
grafik
- Allow dismissing of the toaster, but show it again

- Also log the version number in the backend
- Enable line-breaks with `\n` in toasters

### DIFF
--- a/apps/frontend/src/hooks/useVersionChecker.tsx
+++ b/apps/frontend/src/hooks/useVersionChecker.tsx
@@ -1,0 +1,71 @@
+/*
+ * LICENSE
+ *
+ * This program is free software: you can redistribute it and/or modify it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more details.
+ *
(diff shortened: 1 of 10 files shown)


apps/frontend/src/locales/de/translation.json
michidold, line 1: Personally, I find shorter text looks better.:
grafik

##FFTR

apps/frontend/src/hooks/useVersionChecker.tsx
```diff
@@ -0,0 +1,72 @@
...
+        id: `version-${currentVersion.current}`,
+        duration: Infinity,
+        position: 'top-center',
+        icon: <GrUpgrade />,
+        action: {
+          label: t('version.update'),
+          onClick: () => window.location.reload(),
+        },
+        onDismiss: () => setHasNewVersion(false),
+      });
```
michidold, line 67: Don't know the exact requirement, but does it make sense that we show something, or shall we simply refresh the page?

For the toaster, I would call it as toast.info and position top-right.

```typescript
toast.info(t('version.newVersionAvailable', { appName: APPLICATION_NAME }), {
        id: `version-${currentVersion.current}`,
        duration: Infinity,
        position: 'top-right',
        icon: <GrUpgrade color="green" />,
        action: {
          label: t('version.update'),
          onClick: () => window.location.reload(),
        },
        onDismiss: () => setHasNewVersion(false),
      });
      
 ##FFTR

michidold (approved):
Have some general questions. Maybe we could change the behavior in the future. We should get the version info before logging in. Let's talk about that.

PR #1055, opened 2025-07-10
1054 - FileManager - Add support for edulution-fileproxy

michidold:
- Added transformer function to convert D: to d:
- Added depth 1 to request header
- Added resourcetype to requset body
- Add envs for noroot (in webdav response) and deploymentTarget to differentiate on groups and users requests
- Add a interceptor to check which deploymentTarget
- Renamed GetSchoolPrefix to GetOrganisationsPrefix
- Add global setting entry to configure deploymentTarget from the ui
- Add migration to update global-settings db with deploymentTarget


libs/src/filesharing/utils/parseWebDAVResponse.ts
```diff
@@ -34,6 +34,14 @@ const parseWebDAVResponse = (response: WebdavResponse) => {
     return {} as DirectoryFileDTO;
   }
 
+  const getDisplayName = (name: string | undefined, fallback: string): string | undefined => {
+    if (!name) {
+      const fallbackName = fallback.split('/').filter(Boolean);
```
Maddi02, line 39: Due to plural Fact what about `fallbackNames`

Maddi02 (approved):

PR #1057, opened 2025-07-10
Add french translation

michidold:
Special thanks to @kiarn for providing the translation.

Its not completed yet, but will be in a future task. Therefor the checkTranslation and checkErrorMessage hooks are not updated.

Complete french translation: https://github.com/edulution-io/edulution-ui/issues/1058


TomlDev (commented):
Looks good, but why/how is it not completed yet? Only because the checks are not active?
I would not want to merge it if it's not a complete translation

kiarn:
@TomlDev : do you know which translations are missing ?
There's maybe a little difference between the JSON files, because I've added a key 'fr' in the translations block.

TomlDev (approved):

PR #1073, opened 2025-07-16
1072 - [SurveyCreator] - Error when activating backend limiters (sporadic) 

maierdominik:
Issue: https://github.com/edulution-io/edulution-ui/issues/1072

Image

1072:
* fix error (replace `structuredClone` with `JSON.parse(JSON.stringify())`)
* make sure that the survey formula is defined if not get it from the creator


michidold (approved):
:+1:

PR #1080, opened 2025-07-18
1078 - [SurveyCreator] - Render preview of templates in ResizableWindow, add button to card to show preview

maierdominik:
Issue: https://github.com/edulution-io/edulution-ui/issues/1078

1078: 
* expand the store with the state of the preview dialog
* open the template preview in the parent component to only open it once

ADDITIONALLY: 
* fix the margin for the error messages on the change password form
- got in focus when i tryed to split the button variants like we had to with the width variant of an input -

### DIFF
--- a/apps/frontend/src/pages/Surveys/Editor/SurveyEditorLoadingTemplate.tsx
+++ b/apps/frontend/src/pages/Surveys/Editor/SurveyEditorLoadingTemplate.tsx
@@ -10,7 +10,7 @@
  * You should have received a copy of the GNU Affero General Public License along with this program. If not, see <https://www.gnu.org/licenses/>.
  */
 
-import React, { useState } from 'react';
+import React from 'react';
 import { MdOutlineOpenInNew } from 'react-icons/md';
 import { HiTrash } from 'react-icons/hi';
 import cn from '@libs/common/utils/className';
@@ -22,7 +22,6 @@ import useSurveyEditorPageStore from '@/pages/Surveys/Editor/useSurveyEditorPage
 import useTemplateMenuStore from '@/pages/Surveys/Editor/dialog/useTemplateMenuStore';
 import { Card } from '@/components/shared/Card';
 import { Button } from '@/components/shared/Button';
-import SurveyEditorLoadingPreview from '@/pages/Surveys/Editor/SurveyEditorLoadingPreview';
 
 interface SurveyEditorLoadingTemplateProps {
   creator: AttendeeDto;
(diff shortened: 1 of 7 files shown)


apps/frontend/src/components/shared/Input.tsx
```diff
@@ -102,7 +102,7 @@ const Input = React.forwardRef<HTMLInputElement, InputProps>(
           {...props}
         />
         {type === 'password' ? (
-          <div className="absolute inset-y-0 right-0 flex items-center pr-3 text-sm leading-5">
+          <div className="absolute inset-y-0 right-5 flex items-center pr-3 text-sm leading-5">
```
TomlDev: I am not sure if this is the correct solution, we will talk about it in the daily.

apps/frontend/src/pages/UserSettings/Security/components/PasswordChangeForm.tsx
TomlDev: I would prefer to fix this in another branch.
I wouldn't remove the distance between input fields completely. Also I think it would be nice if the input field would reserve the space for the error. So that if an error is displayed there is no visual down shift of the following components

apps/frontend/src/pages/Surveys/Editor/SurveyEditorLoadingTemplate.tsx
```diff
@@ -64,21 +70,33 @@ const SurveyEditorLoadingTemplate = ({ creator, surveyTemplate }: SurveyEditorLo
...
+        variant="btn-outline"
+        onClick={(e) => {
+          e.stopPropagation();
+          setTemplate(surveyTemplate);
+          setIsOpenTemplatePreview(true);
+        }}
+        className="h-14 w-14 p-2"
+        aria-label={i18n.t('survey.editor.template.preview')}
+      >
+        <MdOutlineOpenInNew className="h-10 w-10" />
```
TomlDev: I think the icon is widely used for "Open in new tab", AI proposes:

> Most common, instantly understood choice is an eye icon:
> Eye → standard “preview/view”

TomlDev (changes requested):
Looks good 👍🏽

TomlDev (approved):

michidold (commented):
Found some visual issues which are necessarily related. Please have a look:
- Cards toggle on hover
- Placeholdertext for searchbar is missing
- On mobile: buttons are cut off
- Delete is somehow not working
- Previewbutton on the bottom
- New survey icon not clickable

Maybe you can fix that in the corresponding branches or create new issues.

https://github.com/user-attachments/assets/d859de44-514a-499c-9681-6d0f4480dac4

michidold (approved):
> Found some visual issues which are necessarily related. Please have a look:
> 
> - Cards toggle on hover
> - Placeholdertext for searchbar is missing
> - On mobile: buttons are cut off
> - Delete is somehow not working
> - Previewbutton on the bottom
> - New survey icon not clickable
> 
> Maybe you can fix that in the corresponding branches or create new issues.
> 
> 
> 
> And it crashes somehow if I return from new survey back to preview:
> 

The issues above are resolved.

PR #1083, opened 2025-07-18
1079 - [SurveyCreator] - Default Logo -> Edulution “Umfragen” Logo

maierdominik:
Issue: https://github.com/edulution-io/edulution-ui/issues/1079

surveys-default-logo-without-bg

STILL MISSING:
* We still need some logic that moves the default logo (visible above) into the detsintation folder (similar for templates)

To test:
- You will have to save the image above to the following path: `data/apps/surveys/defaults/surveys-default-logo-without-bg.png` -

1079:
* create new const holding the default values of a survey
* add a link for the logo that does point to an endpoint were the default logo is served
* add the backend functions to serve the default logo


apps/api/src/surveys/surveys-attachment.service.ts
```diff
@@ -251,17 +254,33 @@ class SurveysAttachmentService implements OnModuleInit {
...
+
+  async serveDefaultIcon(res: Response): Promise<Response> {
+    let defaultIconStream: Response | undefined;
+    try {
+      defaultIconStream = await this.fileSystemService.getResponseWithFileStream(
+        res,
+        join(SURVEYS_DEFAULT_FILES_PATH, SURVEYS_CUSTOM_LOGO),
+      );
+    } catch (error) {
+      // to nothing
```
TomlDev, line 275: I think this produces the error:

Image

Better would be an approach without throwing. First check if the file exists I would say, what do you think?

libs/src/survey/constants/surveys-default-values.ts
```diff
@@ -0,0 +1,30 @@
...
+import i18next from 'i18next';
+import SurveyDto from '@libs/survey/types/api/survey.dto';
+import SurveyFormula from '@libs/survey/types/SurveyFormula';
+import EDU_API_URL from '@libs/common/constants/eduApiUrl';
+import { DEFAULT_FILES, PUBLIC_SURVEYS } from '@libs/survey/constants/surveys-endpoint';
+
+const surveysDefaultValues: Partial<SurveyDto> & { formula: SurveyFormula } = {
+  formula: {
+    title: i18next.t('survey.newTitle').toString(),
+    logo: `${EDU_API_URL}/${PUBLIC_SURVEYS}/${DEFAULT_FILES}`,
```
TomlDev, line 22: "DEFAULT_FILES", shouldn't that be "SURVEYS_LOGO"?

libs/src/survey/constants/initial-survey-form.ts
TomlDev, line 1: When I click on "Neu (Leer)" the logo is empty afterwards, is that on purpose?

Image

TomlDev (changes requested):

PR #1085, opened 2025-07-18
1084 SurveyCreator: Rename category 'text' to `Input field`

maierdominik:
Issue: https://github.com/edulution-io/edulution-ui/issues/1084

1084:
* rename the first option `Eingabefeld` (`Input field`)

image

* _- OLD:_
    * _1084 SurveyCreator: Hide subcategory button in question selection:_
        * _hide the second dropdown -_


TomlDev (commented):
I think we should not do this change for UX reasons.
Better would be to rename the first "Text" to something like "Nutzereingabe" to make it more clear. What do you think @michidold ?
Otherwise when for example "Date" is selected you have no clue that it is selected:

Image

Eingabefeld
Eingabe

TomlDev (approved):
👌🏽

PR #1089, opened 2025-07-23
1076 - Show Totp secret for apps without camera

michidold:


apps/frontend/src/components/ui/QRCodeDisplay.tsx
```diff
@@ -14,13 +14,23 @@ import React, { FC } from 'react';
...
 
 interface QRCodeDisplayProps {
   value: string;
   size?: Sizes;
   className?: string;
+  isLoading?: boolean;
 }
-const QRCodeDisplay: FC<QRCodeDisplayProps> = ({ value, size, className }) => {
+
+const sizeClassMap = {
```
Maddi02, line 26: What about doing something like thaht:

const SIZE_CONFIG = {
  sm: { pixels: 64, class: 'w-[64px] h-[64px]' },
  md: { pixels: 128, class: 'w-[128px] h-[128px]' },
  lg: { pixels: 200, class: 'w-[200px] h-[200px]' },
  default: { pixels: 256, class: 'w-[256px] h-[256px]' },
} as const;

instead of using a switch case

Maddi02 (changes requested):
It works well from my perspective. Only some minor issues.

Maddi02 (approved):

PR #1103, opened 2025-07-29
Fix input buttons

TomlDev:
After some testing I found out it's a problem of using `uuid()` here for the mapped components `key`.

AI explanation:

> Every time you call `uuidv4()` inside your render, you’re handing React a brand‑new key on every single render. React will then:
> 
> 1. **Tear down** the old `<button>` (and its event listeners)
> 2. **Mount** a brand‑new `<button>` in its place
> 
> …all before you ever get a chance to click it. In effect, the node under your cursor is constantly being thrown away and replaced, so your click never “lands” on a stable element that React recognizes.
> 
> When you switch to something stable, like
> 
> ```tsx
> key={ButtonIcon.toString()}
> ```
> 
> the key stays the same across renders, React keeps the same DOM node alive, and your `onClick` gets called as expected.
> 
> ---
> 
> ### How to fix it
> 
> * **If you really need a UUID** (for e.g. server‑generated IDs), generate it *once* outside of render (e.g. in `useMemo` or atop your file) so it doesn’t change every render.
> * Otherwise, use a stable value—like the icon’s own name or your `userAccount.accountId`—for the `key`.
>


michidold (approved):
:+1:

PR #1116, opened 2025-07-31
LDAP Sync: Add LDAP credentials

TomlDev:
grafik


apps/api/src/global-settings/migrations/migration002.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    // eslint-disable-next-line no-underscore-dangle
+    const ids = docs.map((doc) => doc._id);
+
+    const result = await model.updateMany(
+      { _id: { $in: ids } },
+      {
+        $set: {
+          'general.ldap.binduser.dn': '',
+          'general.ldap.binduser.password': '',
```
michidold, line 39: Here the env vars should be loaded. Ether direct or via the defaultValues

michidold (changes requested):

libs/src/common/utils/Object/deepMerge.ts
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+
+type PlainObject = Record<string, unknown>;
+
+const deepMerge = <T>(target: T | null | undefined, source: Partial<T> | null | undefined): T => {
```
michidold, line 15: You could early return if target === source .

##FFTR

michidold (approved):
:+1:

Issue #1102, opened 2025-07-29
Fetch AD Groups via LDAP and Update Keycloak Groups


michidold:
-> See subtasks

Optional:

- [ ] Trigger a debounced full-sync when user updates failed `x` times

PR #1133, opened 2025-08-05
Add LDAP - Keycloak sync to update on group changes

TomlDev:
Adds a periodic sync that fetches the modified groups from LDAP and updates the users in Keycloak.
- Keycloak requests are retried with new token if authentication fails  
- LDAP config is fetched from Keycloak
- LDAP credentials are taken from database
- The service caches the userIds from Keycloak in case the user has multiple new groups
- Currently there are still open issues, see [this parent task](https://github.com/edulution-io/edulution-ui/issues/1102)
  - Requests are currently not written to BullMQ
  - Cache is not updated directly (except `allGroups-global` cache key). So currently one has to wait for they periodically group update of the `GroupsService`

How to test:
1. In ClassManangement you can enroll yourself for a project or class
2. Lookup the user in Keycloak and wait for the delta sync to run (you can decrease the `KEYCLOAK_STARTUP_TIMEOUT`  and `LDAP_SYNC_INTERVAL_MS` values)

For updated `.env` values [see here](https://doku.netzint.de/intern:technical-services:schoolserver:edulution:guides:dev-env#backend).

If you experience issues it can be because of out-of-sync problems, then just flush redis and remove the MongoDB document `ldapkeycloaksyncs` for a full sync and restart the backend (because of the in-memory `lastSync` date).

You need to open an SSH-tunnel like this:
```
ssh -L 1389:10.0.0.1:389   -p 2222   loginuser@ui....

### CODE REVIEW

File: libs/src/ldapKeycloakSync/constants/ldapSyncIntervalMs.ts [hunk 1]
```diff
@@ -0,0 +1,15 @@
...
+ * This program is free software: you can redistribute it and/or modify it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+
+const LDAP_SYNC_INTERVAL_MS = 2 * 60 * 1000;
```
michidold, line 13: 5 mins are sufficient

apps/api/src/scripts/scripts.service.ts
```diff
@@ -22,7 +22,7 @@ class ScriptsService implements OnModuleInit {
 
   @Timeout(60_000)
```
michidold, line 23: Could be replaced by KEYCLOAK_STARTUP_TIMEOUT

apps/api/src/ldap-keycloak-sync/ldap-keycloak-sync.service.ts
```diff
@@ -0,0 +1,327 @@
...
+      Logger.error(
+        `LDAP sync failed: ${(error as Error).message}`,
+        (error as Error).stack,
+        LdapKeycloakSyncService.name,
+      );
+    }
+  }
+
+  private async getBindCredentials(): Promise<{ bindDN: string; bindCredentials: string }> {
+    const bindCredentials = await this.globalSettingsModel.findOne({}, 'general.ldap').lean();
```
michidold, line 151: Normally, we would use the specific service to access the db, so its better to test and maintain. What do you think?

michidold (changes requested):
Have not run it yet, just code review.

michidold (approved):
:+1:  Just one small improvement

Issue #1040, opened 2025-07-08
Survey: Dropdown shows [object Object] as option


TomlDev:
<img width="995" height="856" alt="Image" src="https://github.com/user-attachments/assets/4c929e14-e807-4359-baf3-902772d908c4" />

<img width="720" height="1280" alt="Image" src="https://github.com/user-attachments/assets/ba7a04c1-fcf9-47de-9a51-fa4024aaf3e9" />

PR #1151, opened 2025-08-07
1040 - Survey: Dropdown shows [object Object] as option

maierdominik:
Issue: https://github.com/edulution-io/edulution-ui/issues/1040

1040: 
* error was coused by a wrong rendering of the SHOW_OTHER_ITEM
    * now hidden 
    * default SHOW_OTHER_ITEM entry is prefered


maierdominik:
The problem was the choice `SHOW_OTHER_ITEM` item that is part of the `backendLimiters`,
The item has missed the `title` and could not be displayed accordingly.
But the item that the user selects to add another item is displayed differently,
that is why we filter the option out when displaying the component

PR #1165, opened 2025-08-21
Fix: resolve display issue in proxy configuration settings form

Maddi02:


TomlDev (changes requested):
On some screen sizes it seems to be better, but on small screens its still there...

Small size, it's still there:
Image

Medium size:

https://github.com/user-attachments/assets/7be463e5-4f16-4cf0-829a-88f8967c4f35

TomlDev (approved):
Good job, I tested it and it looks promising. Nonetheless I would like @michidold to check it first before we merge it 👍🏽

PR #1180, opened 2025-08-27
Keycloak Startup: Fix Keycloak-LDAP sync queue

TomlDev:
Same problem like in this fix, just other place.
https://github.com/edulution-io/edulution-ui/commit/170a11fdfef480c0c6519b205bb0eab9dd371f29

We now retry to query Keycloak until it's up for the queue workers.

### DIFF
--- a/apps/api/src/ldap-keycloak-sync/queue/keycloak-request.queue.ts
+++ b/apps/api/src/ldap-keycloak-sync/queue/keycloak-request.queue.ts
@@ -10,17 +10,16 @@
  * You should have received a copy of the GNU Affero General Public License along with this program. If not, see <https://www.gnu.org/licenses/>.
  */
 
-import { Injectable, OnModuleDestroy, OnModuleInit } from '@nestjs/common';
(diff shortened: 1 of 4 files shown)


apps/api/src/ldap-keycloak-sync/queue/keycloak-request.queue.ts
```diff
@@ -77,9 +109,14 @@ export default class KeycloakRequestQueue implements OnModuleInit, OnModuleDestr
...
     const job = await this.queue.add(
-      QUEUE_NAME,
+      QUEUE_CONSTANTS.KEYCLOAK_REQUESTS_QUEUE,
       { method, endpoint, payload, config },
-      { removeOnComplete: true, removeOnFail: true, attempts: 3 },
+      {
+        removeOnComplete: true,
+        removeOnFail: true,
+        attempts: 10,
+        backoff: { type: 'exponential', delay: 3000 },
```
Maddi02, line 118: Could we use some constants in here?

Maddi02 (approved):
👌🏼. Perhaps we could use constants instead of 'magic strings'.

PR #1191, opened 2025-09-01
LDAP Sync: Clean up console logging

TomlDev:
When the binduser was not correctly configured in the ENV, the LDAP sync service logged an error. This is unnecessary because it could be that it's only configured in the DB.


apps/api/src/ldap-keycloak-sync/ldap-keycloak-sync.service.ts
```diff
@@ -83,7 +77,7 @@ class LdapKeycloakSyncService implements OnModuleInit {
   @Interval(LDAP_SYNC_INTERVAL_MS)
   async sync() {
     if (!this.ldapConfig) {
-      Logger.error('No LDAP config', LdapKeycloakSyncService.name);
+      Logger.error('No LDAP config, sync canceled', LdapKeycloakSyncService.name);
```
michidold, line 80: Is it an error or just a warning? Think it could be on purpose. What do you think?

michidold (commented):
Just two questions

PR #1202, opened 2025-09-03
[Filesharing] Add a pdf renderer as a fallback if OnlyOffice is not configured

Maddi02:
- A PDF renderer has been added that uses the PDF rendering engine from the browser.


TomlDev (approved):
Minor improvements possible 👍

PR #1204, opened 2025-09-03
Reduce maximum file upload size from 400 MB to 50 MB

Maddi02:
To prevent LMN from crashing, the upload limit has been reduced back to 50 MB.


michidold (approved):
:+1:

PR #1216, opened 2025-09-08
1126 - [SurveyTemplate] Add the following default templates

maierdominik:
Issue: https://github.com/edulution-io/edulution-ui/issues/1126

1126:
* fix templates, testing backendLimiters and such
* improve the default templates (wording)]
* move to new branch after the changes from 1142 seemed to be included even so 1142 was the base branch (confusing)

### DIFF
--- a/apps/api/src/surveys/surveys-template.service.ts
+++ b/apps/api/src/surveys/surveys-template.service.ts
@@ -19,12 +19,13 @@ import { HttpStatus, Injectable, OnModuleInit } from '@nestjs/common';
 import CommonErrorMessages from '@libs/common/constants/common-error-messages';
 import getCurrentDateTimeString from '@libs/common/utils/Date/getCurrentDateTimeString';
 import SURVEY_TEMPLATES_EXCHANGE_PATH from '@libs/survey/constants/surveyTemplatesExchangePath';
-import SURVEY_TEMPLATES_DEFAULT_TEMPLATE_PATH from '@libs/survey/constants/surveyTemplatesDefaultTemplatePath';
 import { SurveyTemplateDto, TemplateDto } from '@libs/survey/types/api/surveyTemplate.dto';
 import getIsAdmin from '@libs/user/utils/getIsAdmin';
 import { SurveysTemplate, SurveysTemplateDocument } from 'apps/api/src/surveys/surveys-template.schema';
 import CustomHttpException from '../common/CustomHttpException';
 import FilesystemService from '../filesystem/filesystem.service';
+import MigrationService from '../migration/migration.service';
+import surveysTemplateMigrationsList from './migrations/surveysTemplateMigrationsList';
 
 @Injectable()
 class SurveysTemplateService implements OnModuleInit {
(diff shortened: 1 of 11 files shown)


maierdominik:
Default templates should not be deleteable when a default template should be deleted
set it to `isActive = false` instead.

apps/api/src/surveys/surveys-template.service.ts
```diff
@@ -17,27 +17,32 @@
...
+    @InjectConnection() private readonly connection: Connection,
     @InjectModel(SurveysTemplate.name) private surveyTemplateModel: Model<SurveysTemplateDocument>,
     private readonly globalSettingsService: GlobalSettingsService,
   ) {}
 
   async onModuleInit() {
+    const collectionsNamedMigration = await this.connection.db?.listCollections({ name: 'surveystemplates' }).toArray();
+    if (collectionsNamedMigration?.length === 0) {
+      await this.connection.db?.createCollection('surveystemplates');
+    }
```
TomlDev: I think this is not necessary. As far as I know, MongoDB will create new collections when necessary automatically

libs/src/survey/constants/templates/Elternbrief.ts
```diff
@@ -0,0 +1,104 @@
...
+ *
+ * If you are uncertain which license applies to your use case, please contact us at info@netzint.de for clarification.
+ */
+
+const eduHost = process.env['EDUI_HOST'] ?? 'localhost';
+const eduPort = process.env['EDUI_PORT'] ?? '3001';
+const frontendHost = process.env['FRONTEND_HOST'] ?? 'localhost';
+const frontendPort = process.env['FRONTEND_PORT'] ?? '5173';
+
+const Elternbrief = {
```
TomlDev: Please use our normal naming schema (english, camelCase, etc.) for all of the templates

apps/api/src/surveys/migrations/surveyTemplatesMigration001LoadDefaultTemplates.ts
```diff
@@ -0,0 +1,59 @@
...
+const name = '001-load-the-default-survey-templates';
+
+const surveyTemplatesMigration001LoadDefaultTemplates: Migration<SurveysTemplateDocument> = {
+  name,
+  version: 1,
+  execute: async (model) => {
+    Logger.log(`Migration "${name}": Found ${list.length} documents to process...`);
+    await Promise.all(
+      list.map(async (surveyTemplate) => {
+        const existingTemplate = await model.findOne({ name: surveyTemplate.name, isDefaultTemplate: true });
```
TomlDev: I think we should use IDs here instead of names

libs/src/survey/constants/templates/000000000000000000000004-parentTeacherConference.ts
```diff
@@ -0,0 +1,278 @@
...
+ *
+ * 2. A commercial license agreement with Netzint GmbH. Licensees holding a valid commercial license from Netzint GmbH
+ *    may use this software in accordance with the terms contained in such written agreement, without the obligations imposed by the AGPL.
+ *
+ * If you are uncertain which license applies to your use case, please contact us at info@netzint.de for clarification.
+ */
+
+import { Types } from 'mongoose';
+
+const ParentTeacherConference = {
```
TomlDev: How could you push this? It shouldn't be allowed by naming conventions

TomlDev: alright, maybe we can switch number and name?
`letterToParents_000000000001`

michidold (approved):
Found some improvements. Pushed them afterwards.

PR #1238, opened 2025-09-16
Handle whitespace trimming and URL encoding for file and folder operations

Maddi02:
- Added functions to remove whitespace at the end of a folder name.
- Ensured that the encoding works for every operation.


michidold (changes requested):
Works not reliably, some times it has wrong encoding. Think that whole path handle is too complicated. We need to handle it only once.

Issue #1242, opened 2025-09-17
[Survey] - File question secure the endpoint to retrieve answered files


maierdominik:
Move @Get(`${ANSWER}/${FILES}/:userName/:surveyId/:filename`) out of the public and into the secure controller.
<img width="656" height="355" alt="Image" src="https://github.com/user-attachments/assets/118da32e-6b6f-47d2-8f75-e8f13797fb59" />

Add a check if the downloading user is the creator of the survey, only he will be allowed to download the files.

PR #1249, opened 2025-09-17
1242 - [Survey] - File question secure downloading endpoint

maierdominik:
Parent issue: https://github.com/edulution-io/edulution-ui/issues/1222

issue: https://github.com/edulution-io/edulution-ui/issues/1242

1242:
* split the file serving into an temporary (public-) and a permanent (private-) option
* when moving the answers attachments we have to update the url ( by updating the `.content` property ) to link to the new protected endpoint
 * add the check if the person that fetches the files is the creator of the survey


TomlDev:
I hope you tested it a lot!

PR #1257, opened 2025-09-18
Update vulnerable multer package

TomlDev:
Updated `multer` to v2, added tests where used to check for correct vulnerability handling.


michidold (approved):
:+1:

PR #1264, opened 2025-09-19
ClassManagement: Fix bugs with printing passwords as admin

TomlDev:
- Allow multiple selection for printing passwords only for the same school (otherwise it fails)
- Send the `school` correctly
- Show a loading indicator when loading printable classes
- Show all classes for admins
- Fix the file type of the downloaded CSV (before the browser interpreted it as PDF)


michidold (approved):
:+1:

PR #1270, opened 2025-09-19
[Filesharing] Add the option to download a folder or multiple files

Maddi02:
- Added logic to stream/download folders.
- Download multiple files without buffering them in the backend. Now multiple file streams are open.


michidold (changes requested):
Found one improvement. Please have a look

PR #1282, opened 2025-09-24
Enhance download progress tracking

Maddi02:
- Extended the download progress information (speed, estimated time of arrival and bytes) in the DownloadFileDto.
- Implemented live calculation of speed/ETA during onDownloadProgress.
- Fixed the bug where the download toast remained even after the download had finished. 

https://github.com/user-attachments/assets/6707b2cb-3658-4171-9495-a4eaa48f3d37


apps/frontend/src/pages/FileSharing/useFileSharingDownloadStore.ts
```diff
@@ -123,29 +125,80 @@ const useFileSharingDownloadStore = create<FileSharingDownloadStore>((set, get)
...
+              speedBps: smoothedBytesPerSecond,
+              etaSeconds: estimatedSecondsRemaining,
+              lastUpdateAt: currentTimestampMilliseconds,
+              startedAt,
+              speedFormatted,
+              etaFormatted,
             });
+
+            lastTimestampMilliseconds = currentTimestampMilliseconds;
+            lastLoadedBytes = loadedBytesResolved;
```
TomlDev, line 201: Maybe take a look on the AI suggestions (I don't know if any of them are applicable: 
<details>
Here are safe refactors that **don’t change behavior** (inputs/outputs stay the same):

### What I’d tweak (no logic changes)

* Use a clearer flag name: you toggle `isFetchingPublicUrl` inside `downloadFile`. Rename to something generic like `isDownloading` (keeps semantics correct). If you can’t rename globally, leave it as-is.
* Keep return type consistent: you return `string | undefined` in the type but return `''` on errors. Either return `undefined` or keep `''` consistently. Below I keep your existing behavior (returning `''` on error).
* Extract tiny helpers to reduce inline math/noise.
* Prefer `performance.now()` for better timing resolution (still fine to keep `Date.now()`, I’ll keep it to avoid logic drift).
* (Optional) expose a `revokeObjectURL` responsibility comment so callers don’t leak memory.

### Drop-in refactor (same logic/behavior)

```ts
downloadFile: async (file, signal) => {
  try {
    // Keep your current flag to avoid wider changes
    set({ isFetchingPublicUrl: true, error: null });

    const totalBytesFallback = file.size ?? 0;
    const processId = Math.floor(Math.random() * 1_000_000);
    const startedAt = Date.now();

    // Progress state vars
    let lastTs = startedAt;
    let lastLoaded = 0;
    let smoothedBps: number | undefined;

    const formatFileName = () =>
      file.type === ContentType.DIRECTORY ? `${file.filename}.zip` : file.filename;

    // Initialize progress
    get().setDownloadProgress({
      fileName: formatFileName(),
      percent: 0,
      processId,
      totalBytes: totalBytesFallback,
      loadedBytes: 0,
      speedBps: 0,
      etaSeconds: undefined,
      lastUpdateAt: startedAt,
      startedAt,
      speedFormatted: formatTransferSpeed(0),
      etaFormatted: formatEstimatedTimeRemaining(undefined),
    });

    const onProgress = (pe: AxiosProgressEvent) => {
      const total = pe.total ?? totalBytesFallback;
      if (!total) return; // unknown total: skip

      const loaded = pe.loaded ?? 0;
      const currentTs = Date.now();

      // percent (clamped)
      let percent = Math.round((loaded / total) * 100);
      if (percent > 100) percent = 100;

      // instantaneous speed
      const dtMs = Math.max(1, currentTs - lastTs);
      const dBytes = Math.max(0, loaded - lastLoaded);
      const instBps = (dBytes * 1000) / dtMs;

      // exponential smoothing (same factor)
      const alpha = 0.2;
      smoothedBps = smoothedBps ? alpha * instBps + (1 - alpha) * smoothedBps : instBps;

      // ETA
      const remaining = Math.max(0, total - loaded);
      const etaSec = smoothedBps > 0 ? remaining / smoothedBps : undefined;

      get().setDownloadProgress({
        fileName: formatFileName(),
        percent,
        processId,
        totalBytes: total,
        loadedBytes: loaded,
        speedBps: smoothedBps,
        etaSeconds: etaSec,
        lastUpdateAt: currentTs,
        startedAt,
        speedFormatted: formatTransferSpeed(smoothedBps),
        etaFormatted: formatEstimatedTimeRemaining(etaSec),
      });

      lastTs = currentTs;
      lastLoaded = loaded;
    };

    const { data } = await eduApi.get<Blob>(
      `${FileSharingApiEndpoints.FILESHARING_ACTIONS}/${FileSharingApiEndpoints.FILE_STREAM}`,
      {
        responseType: ResponseType.BLOB,
        signal,
        params: { filePath: file.filePath },
        onDownloadProgress: onProgress,
      },
    );

    // NOTE: caller should eventually call URL.revokeObjectURL(url) after use
    return URL.createObjectURL(data);
  } catch (error) {
    handleApiError(error, set);
    return '';
  } finally {
    set({ isFetchingPublicUrl: false });
  }
},
```

### Minor consistency suggestions (optional)

* **Return undefined on error** everywhere (including `createDownloadBlobUrl` and `getPublicDownloadUrl`) instead of `''`, and reflect that in the store types. This avoids “empty string means failure” checks.
* Consider a tiny shared fetch helper to DRY up calls to the same endpoint used by `createDownloadBlobUrl` and `downloadFile` (same URL, similar config) while keeping `onDownloadProgress` optional.
* If you can, rename `isFetchingPublicUrl` → `isDownloading` in this slice; it’s semantically clearer, but it’s cosmetic.

If you want, I can show a small shared `fetchFileBlob({ filePath, signal, onProgress })` helper and plug both `createDownloadBlobUrl` and `downloadFile` into it—no logic change, just de-duplication.
</details>

TomlDev (approved):

michidold (approved):
Looks good. Don't forget the share prop when resolving the conflict.

PR #1287, opened 2025-09-25
ClassManagement: Fix bugs on lesson page

TomlDev:
There is a problem with nested groups in LMN. The LMN api only responds with the correct `member` array, if you query it with a special attribute. The problem here is that these requests take a very long time.

- This PR fixes a problem where the LMN api changed (`sophomorixMembers` is no more valid, had to be replaced with `all_members`.
- Additionally we now show a plus `+` when there are nested groups as members of groups when we expect more students in a group:
grafik


michidold (approved):
👍

PR #1291, opened 2025-09-25
AuthGuard: Log only in debug mode

TomlDev:
The idea is to log client errors (http status 4xx) in production only in debug mode (e.g. UNAUTHORIZED) to clean up the logs a little.


michidold (approved):
👍

PR #1297, opened 2025-09-26
Filesharing: Add dynamic user paths

michidold:


apps/api/src/webdav/shares/webdav-shares.schema.ts
TomlDev, line 1: Not related: Missing translations

Image

TomlDev, line 1: For me the `webdavShare.variable` dropdown is not clickable (there are no options) 

Image

TomlDev (changes requested):
Code looks good so far but I could not test the functionality

TomlDev (approved):

TomlDev:
Please add the missing translations and add a info to the user that no rootShare is configured.
grafik

TomlDev (approved):

PR #1298, opened 2025-09-26
[Filesharing] Fix the incorrect success message when creating a file

Maddi02:
https://github.com/user-attachments/assets/b0e03251-d5ce-486b-bd10-18477ac30a76

https://github.com/user-attachments/assets/85909071-b559-4187-8db2-c909f28c5acc


apps/api/src/webdav/webdav.service.ts
```diff
@@ -245,34 +246,32 @@ class WebdavService {
...
+      });
+      fileStream.on('aborted', () => request.cancel());
+      fileStream.on('error', () => request.cancel());
+
+      void request.on('uploadProgress', (p) => onProgress?.(p.transferred, p.total));
+
+      const response = await request;
+      const ok = response.statusCode >= 200 && response.statusCode < 300;
+      return { success: ok, status: response.statusCode, filename: fullPath.split('/').pop() || '' };
+    } catch (error) {
```
michidold, line 272: If the upload is aborted, by eg refreshing the page, the same error is triggered. Can you add a task, where you add a abort button to the toast and add here a handler if the upload is aborted?

apps/frontend/src/hooks/useUploadProgressToast.tsx
michidold, line 1: This hook is now only executed if you stay on the filesharing page. I would move it to the `GlobalHooksWrapper`. Else it freezes while navigating to other apps.

michidold (changes requested):

apps/api/src/webdav/webdav.service.ts
Maddi02: #1308

michidold (approved):

PR #1321, opened 2025-10-06
Fix: After uploading a file, the file list cannot be refreshed.

Maddi02:
- Fixed upload params (`name`, `share`, `contentLength`) for WebDAV uploads.
- Added progress tracking per `share` and file `id`.

### DIFF
--- a/apps/frontend/src/pages/FileSharing/utilities/UploadContentBody.tsx
+++ b/apps/frontend/src/pages/FileSharing/utilities/UploadContentBody.tsx
@@ -37,6 +37,7 @@ import { FcFolder } from 'react-icons/fc';
 import MAX_FOLDER_UPLOAD_CONTENT_SIZE from '@libs/ui/constants/maxFolderUploadContentSize';
 import getFileUploadLimit from '@libs/ui/utils/getFileUploadLimit';
 import useHandelUploadFileStore from '@/pages/FileSharing/Dialog/upload/useHandelUploadFileStore';
+import { v4 as uuidv4 } from 'uuid';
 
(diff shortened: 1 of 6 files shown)


apps/frontend/src/pages/FileSharing/utilities/UploadContentBody.tsx
```diff
@@ -123,8 +126,21 @@ const UploadContentBody = () => {
       }
 
       updateFilesToUpload((prevFiles) => {
-        const allNewFiles = acceptedFiles.filter((file) => !prevFiles.some((f) => f.name === file.name));
-        return [...prevFiles, ...allNewFiles];
+        const existingNames = new Set(prevFiles.map((f) => f.name));
+
+        const newFiles = acceptedFiles
```
TomlDev, line 131: Shouldn't here the ID be checked instead of name?

apps/frontend/src/pages/FileSharing/utilities/UploadContentBody.tsx
```diff
@@ -123,8 +126,21 @@ const UploadContentBody = () => {
...
+          .filter((file) => !existingNames.has(file.name))
+          .map((file) => {
+            const uploadFile: UploadFile = Object.assign(new File([file], file.name, { type: file.type }), {
+              id: uuidv4(),
+              isZippedFolder: false,
+              originalFolderName: undefined,
+              fileCount: undefined,
```
TomlDev, line 138: Is it necessary to make it undefined? Instead of making the prop optional and leaving it out?

TomlDev (dismissed):

